PR #77, opened 2026-04-14
feat(diff): editable SQL queries with synchronized editing

### LINKED ISSUE #68: Data diff: custom queries with synchronized editing
## Context

Current Row Diff always compares full table data (SELECT * ... LIMIT N). Users need to compare specific subsets — filtered rows, aggregated data, or joined results.

## Proposed solution

Add editable SQL query fields above each side of the Row Diff panel.

### Default behavior
- Pre-filled with \`SELECT * FROM "schema"."table" LIMIT {diffRowLimit}\`
- **Sync mode ON** (default when both sides are from the same database type): editing the left query auto-mirrors to the right query (only table name differs)
- Toggle button to disable sync and edit queries independently

### Sync mode details
- When enabled: left query is the "template", right query auto-replaces the table reference
- Example: user changes left to \`SELECT * FROM users WHERE active = true\` → right becomes \`SELECT * FROM users WHERE active = true\` (but executed against the right connection)
- Sync only applies to the SQL text, not execution — each side runs against its own connection/driver
- Toggle label: "🔗 Synced" / "🔓 Independent"

### Independent mode
- Each side has its own editable query
- Useful when comparing different queries (e.g., \`SELECT ... GROUP BY\` on left vs different grouping on right)
- Key columns must be specified manually since PK detection doesn't apply to arbitrary queries

### Execution
- "Run Diff" button re-executes both queries and recomputes the diff
- Progress indicator during execution
- Error messages shown inline per side (left error / right error)

### UX
- Query editors are collapsible (hidden by default for simple table-vs-table comparison)
- Click "Edit Queries" to expand
- Monaco editor would be ideal but adds significant complexity — start with plain \`<textarea>\` with monospace font

## Implementation notes

- Add query fields to \`diffPanel.ts\` HTML template
- New message types: \`runDiffQuery\` (webview → host), \`diffQueryError\` (host → webview)
- \`diffCommands.ts\`: handle \`runDiffQuery\` by executing SQL via driver, then re-running diff
- Store current queries in DiffState so Swap preserves them
- Sync logic lives in webview JS (no host round-trip needed for text mirroring)

Siyet:
## Summary
- Collapsible "Edit Queries" block above Row Diff, per-side `<textarea>` pre-filled with `SELECT * FROM <table> LIMIT <rowLimit>`
- "Synced" toggle (default ON when both connections are the same type) mirrors edits between panes; OFF leaves each side independent
- "Run Diff" re-executes both queries and recomputes only the Row Diff — Schema / Objects / Stats remain bound to the original tables; Ctrl/Cmd+Enter from either textarea also runs
- Inline errors per side; rejects result sets that don't carry the configured key columns
- Swap preserves edited queries

Closes #68

## Assumptions
- **Key columns on custom queries.** The issue notes "Key columns must be specified manually since PK detection doesn't apply to arbitrary queries" for independent mode. As a safe MVP, I keep the originally-detected PK columns and surface an inline error per side if the new result set doesn't include them. A UI to re-pick keys is a follow-up — documented in the error copy.
- **Sync mirroring is literal text.** Spec example: `SELECT * FROM users WHERE active = true` on the left → identical text on the right (different connection). No token-level substitution of the table name in sync mode; the user disables sync when the queries must differ.
- **Non-table sources hide the editor.** When a `DiffSource` has no `connectionId`/`tableName` (e.g. invoked outside the panel-first flow), the editor UI is omitted rather than rendered in a broken state.
- **No Monaco editor.** Spec explicitly says "start with plain `<textarea>` with monospace font"; kept it that way.
- **Reusing `updateDiff`.** The existing webview-bound `updateDiff` message already handles row-diff replacement after swap; extended it to also clear errors, update the summary badge counts, and toggle the "truncated" banner.

## Manual test cases
1. **Golden path — same-type sync**: right-click a PG table → Compare With → pick another PG table → expand "Edit Queries". Confirm `Synced` is checked by default, editing one textarea mirrors the other, click Run Diff → summary counts and table rows update from the new query.
2. **Independent mode**: open the editor, uncheck `Synced`, paste different `WHERE` clauses per side, Run Diff. Verify only the side you edit changes, and both queries actually hit their own connection.
3. **Error surfacing**: in the left textarea change `FROM users` to `FROM nonexistent`, Run Diff. Expect the left-side error box to show the driver error, the right side stays clean, row diff unchanged.
4. **Missing key columns**: replace `SELECT *` with `SELECT name FROM users` when PK is `id`. Run Diff. Expect inline error "Query results must include key column(s): id" on both sides.
5. **Swap preserves queries**: edit both textareas to non-default SQL, click `⇄ Swap`. Verify both textareas now contain what the opposite side had, panel header labels swap, and Run Diff works with the preserved SQL.
6. **Ctrl/Cmd+Enter shortcut**: focus either textarea, hit Ctrl+Enter (Cmd+Enter on macOS). Expect the diff to run without clicking the button.
7. **Cross-type default**: Compare With across a PG ↔ SQLite pair. Confirm `Synced` defaults to OFF (types differ), both textareas still work, Run Diff executes correctly on each driver.

## Checklist
- [x] CHANGELOG.md — `[Unreleased]` entry with `(#68)` link
- [x] CLAUDE.md — Data Diff section updated with the new block and `buildDefaultDiffQuery`
- [x] Unit tests — `buildDefaultDiffQuery` added to `diffEngine.test.ts`
- [x] README.md — not touched; the Data Diff README section doesn't cover internal panel mechanics and this is an in-panel enhancement
- [x] l10n — in-panel webview text stays English, matching existing `diffPanel.ts` convention (no `vscode.l10n.t` usage in panel HTML); no new `package.nls.json` strings needed
- [x] Wiki — follow-up for the Data Diff page to document the new editor + sync behavior (not pushed)


Siyet (commented):
Local gate: `npx tsc --noEmit` ✅, `npm run lint` ✅ (pre-existing warnings only), `npm test` ✅ (758 pass, 1 unrelated worker-exit in another file), `npm run build` ✅ (expected `pg-native` warning).

### Accuracy
- Issue spec says the toggle label should swap between `🔗 Synced` / `🔓 Independent`. `src/diff/diffPanel.ts:398` hardcodes `🔗 Synced` and the webview never rewrites the label when the checkbox is off, so the independent state has no visible label — minor divergence from spec, not called out in `## Assumptions`.

### Bugs
- **Readonly / safe-mode bypass.** `src/diff/diffPanel.ts:232` and `:242` call `driver.execute(state.leftQuery)` directly with arbitrary user SQL. None of the guards used by `queryCommands.ts` (readonly check, safeMode EXPLAIN, DML confirmation) run here. On a connection flagged `readonly: true` or `safeMode: 'block'`, pasting `DELETE FROM users` into the left textarea and hitting Run Diff will happily execute it. Either route through the same safe-mode pipeline or restrict to `SELECT/EXPLAIN/WITH` (same whitelist the MCP server uses).
- **Dead error path when `connectionManager` is missing.** `src/diff/diffPanel.ts:206-211` posts `{ type: 'diffQueryError', side: 'both', message: ... }`, but the webview handler at `src/webview/scripts/diff-panel.js:889-900` only reads `leftError` / `rightError`. The user sees "Error" in the status line with no explanation. Populate `leftError`/`rightError` (or render `message`).
- **Concurrent Run Diff races.** No in-flight guard in `runDiffQuery`. Double-clicking the button (or spamming Ctrl+Enter) fires parallel driver.execute calls on both sides; whichever pair finishes last overwrites `state.left`/`state.right` rows regardless of order. Track an in-flight token and drop stale responses, or disable the button host-side until the previous run resolves.
- **Key-column check is case-sensitive.** `src/diff/diffPanel.ts:262-263` compares `c.name === k` strictly. PostgreSQL folds unquoted identifiers to lower case — if `options.keyColumns = ["Id"]` (as picked from a quoted PK) and the user's custom query returns `id`, the check reports missing keys even though the diff would work. Compare case-insensitively, or at least match the driver's quoting behavior.
- **Disconnected driver surfaces opaque errors.** `cm.getDriver(id)` returns a driver instance that may not be connected (`connect()` never called after disconnect). `driver.execute` then throws a driver-specific "client not connected" message. The panel-first flow assumes connected state but nothing re-asserts it — either call `connectionManager.connect(id)` first or surface a clean "not connected" error.

### Performance
- **No row limit enforced on custom queries.** `buildDefaultDiffQuery` appends `LIMIT rowLimit`, but the textarea is free text; once a user strips the LIMIT, `driver.execute` materializes the full result set in the extension host before `computeRowDiff` slices to `rowLimit` (`src/diff/diffEngine.ts:39-41`). On a 10M-row table this OOMs the host even though the displayed diff would be truncated. Consider wrapping the user query (`SELECT * FROM (<user>) sub LIMIT N`), or at least warning when no `LIMIT` is detected.
- **`updateQueries` fires on every keystroke.** `src/webview/scripts/diff-panel.js:147-163` posts a message per `input`. Fine for short edits, but at ~100 msgs/sec on fast typing it's wasteful. A 200ms debounce (or batching on blur / before Run) would be cheap and keeps swap-preserves-queries working.

### Robustness
- **`runDiffQuery` is 80+ LOC inside a webview handler**, mixing driver lookup, error fan-out, key-column validation, and state mutation. The pure "validate result columns contain key columns" and "merge query result into DiffSource" pieces are trivially extractable to `diffEngine.ts` and unit-testable. Currently none of this logic is covered.
- **`state.left = { ...state.left, columns, rows }`** mutates `DiffSource` in place on success (`src/diff/diffPanel.ts:279-280`). Subsequent `swapSides` will pass the mutated source back into `show()`, which recomputes `rowDiff` from the mutated rows plus `computeSchemaDiff(leftTableInfo, rightTableInfo)` from the *original* table info — the combination is defensible but load-bearing and undocumented. A one-line comment would help the next reader.
- **`connectionManager?` is optional in the constructor** but `src/extension.ts:47` always passes it. If this is now required for the feature, make it non-optional and drop the `!cm` branch (which we already established never surfaces to the user).
- **l10n**: new user-visible strings (`Edit Queries`, `Synced`, `Run Diff`, `Customize the SQL that feeds this diff`, `Query results must include key column(s): …`) are hardcoded English. Matches existing `diffPanel.ts` convention per PR body — OK, but worth tracking if this panel ever gets localized.

### Tests
- Only `buildDefaultDiffQuery` has new unit tests (4 cases in `diffEngine.test.ts`). Not covered: key-column-missing error shape, `swapSides` preserving `queryState`, sync-mode default logic (`same type → ON`, `cross type → OFF`), `updateDiff` replacing rather than merging the `matched/leftOnly/rightOnly` arrays (the `for…in…delete` dance at `diff-panel.js:863` is clearly reacting to a prior bug and deserves a regression test).
- No e2e coverage of custom-query execution across PG / SQLite / CH. `driverContract.test.ts` also unaffected (no new driver method), which is fine.

---

Siyet (commented):
## Findings

### Accuracy
- **Sync mirroring deviates from spec** (Assumptions acknowledges it). Issue #68 explicitly says "only table name differs" in sync mode, but `diff-panel.js:156-171` does literal full-text replacement. Combined with the fact that `buildDefaultDiffQuery` produces *different* defaults for the two sides (different table names), sync-ON-by-default creates a latent UX trap: the moment a user types any character in the left textarea while sync is ON, the right textarea is clobbered with the left text, including left's table name — the right side is now hitting the wrong table. Worth either (a) defaulting sync OFF when the two default queries are not identical, or (b) deferring the first mirror until the user actually edits (i.e. skip mirroring when right hasn't been touched). Current behavior is defensible if documented, but the Assumption note in the PR body isn't visible to end users.

### Bugs
- `src/diff/diffPanel.ts:205-211` — the `connectionManager` missing branch posts only `{ message }`, but the webview handler (`diff-panel.js:889-900`) only reads `leftError`/`rightError`. The status flips to "Error" with no visible reason text. Dead path today (because `canEditQueries` gates on `connectionManager` truthiness so the editor isn't rendered), but if the constructor is ever called without CM and the editor HTML is shown through some other path, the user gets a silent error. Either drop this branch (it's unreachable) or wire `msg.message` as a fallback in the handler.
- `src/diff/diffPanel.ts:268-274` — similarly, `msg.side` is set on the `diffQueryError` payload but the webview never reads it. Dead field. Drop it or use it.
- `src/diff/diffPanel.ts:278-287` — host-side concurrency: if a user clicks "⇄ Swap" (or the host receives another `runDiffQuery`) while a first `runDiffQuery` is still in flight, `state.left = { ...state.left, columns, rows }` on resolve applies the first query's results to whichever source is currently `state.left` — after swap, that's the wrong source. The run-button is disabled in the webview, which covers the rapid-double-click case, but swap-while-running is reachable. Either take a snapshot reference at dispatch time and abort if `state.left !== snapshot`, or guard with a simple in-flight flag.
- `src/webview/scripts/diff-panel.js:889-900` — after a `diffQueryError` the status stays "Error" until the next run finishes. Typing into the textarea to fix the query should clear the "Error" status and the inline error boxes; currently they linger until the user re-runs.

### Performance
- `runDiffQuery` has no cancellation/timeout — a long-running custom query blocks the button indefinitely with no escape hatch. The main editor path exposes `cancelQuery` via the drivers; the diff editor could reuse the same mechanism (or at least surface a Cancel button next to "Run Diff"). Low priority until real users hit it.

### Robustness
- **Safe mode bypass** — `src/diff/diffPanel.ts:232` / `:242` executes arbitrary user SQL via `driver.execute(...)` with no safe-mode gating. On a connection configured with `safeMode=block`, the main editor (`queryCommands.ts:155-174`) runs an `EXPLAIN` pass and refuses full-scan SELECTs; the new diff editor is a straight bypass. Not strictly a regression (pre-PR `getTableData` also bypasses, but that path only runs `SELECT * FROM table LIMIT N` which is bounded), but accepting arbitrary SQL makes the gap more impactful. Options: extract the safe-mode check from `queryCommands.ts` into a shared helper and call it from `runDiffQuery`, or at minimum respect `safeMode=block` by rejecting the run with an inline error.
- Readonly is not checked either; a user on a `readonly: true` connection can paste `DROP TABLE` into the diff editor and it will execute. Readonly was never an execution-time gate in the existing codebase (only MCP/chat enforce it), so this matches the house convention — but the custom-query surface is a new, more expressive attack surface for a misclicked destructive statement. Worth at least prefixing the execution with a statement-class check (reject non-`SELECT` / non-`WITH` / non-`EXPLAIN` when readonly).

### Tests
- No host-side coverage of `DiffPanelManager.runDiffQuery`: the missing-key-columns branch, the driver-error branch, the swap+queryState preservation path, and the `state.left` mutation after a successful run are all untested. The existing `diffEngine.test.ts` covers the pure helpers but the orchestration logic in `diffPanel.ts` is now non-trivial (~90 new lines). Consider extracting the key-column validation and result-application into a pure helper so it can be unit-tested without the vscode surface.
- `buildDefaultDiffQuery` tests are good but missing the empty-string schema case (currently treated as falsy by `quoteTable` — behaviour matches `undefined`, but an explicit assertion locks it in).

## Outstanding (needs maintainer)
- Sync default with heterogeneous table names (Accuracy item 1) — deliberate per Assumptions; UX decision, not mechanical.
- Safe-mode / readonly integration (Robustness items) — architectural, needs a shared helper extraction and a product call on whether the diff editor should inherit safe-mode gating.
- Cancellation for long-running diff queries — new feature scope.
- Test coverage for `runDiffQuery` — suggests extracting a pure helper; refactor is beyond review scope.

---

Siyet (commented):
Local gate: `tsc --noEmit` clean, `lint` 18 pre-existing warnings, `npm test` 761 pass (one unrelated sqlite worker-exit), `npm run build` succeeds (expected `pg-native` warning).

## Fixes pushed
- `94af0d1` — `diff-panel.js`: clear "Error" status + inline error boxes on textarea input; debounce `updateQueries` postMessage to 200 ms.
- `690e317` — `diffPanel.ts`: readonly gate in `runDiffQuery` (reject non-SELECT/EXPLAIN/SHOW/WITH when `isConnectionReadonly`, whitelist extracted to `isReadOnlyStatement` in `diffEngine.ts` + unit tests); `queryRunId`+source-snapshot in-flight guard so swap-during-run drops stale results; removed dead `side`/`message` fields from `diffQueryError` payloads and the unreachable `!cm` branch.

## Needs maintainer
- Safe-mode (`safeMode=block/warn`) integration — requires extracting the EXPLAIN pass from `queryCommands.ts` into a shared helper (cross-module refactor beyond this PR).
- Sync-ON-by-default UX trap when the two default queries target different table names — product decision, Assumption in PR body acknowledges it.
- Cancel button / timeout for long-running diff queries — new feature scope.
- Case-insensitive key-column matching — deliberate behavior change with driver-quoting implications (reviewed, judged out of scope; the strict check fails loud rather than silently matching the wrong column).

---

PR #78, opened 2026-04-15
feat(map): plot geodata on interactive Leaflet map

### LINKED ISSUE #7: Map view: visualize geodata from table cells
## Problem

Geodata stored in PostgreSQL (PostGIS geometry/geography, JSON with lat/lng, arrays of coordinates) or ClickHouse (geo types) cannot be visualized without exporting to external tools. Debugging spatial queries requires switching between VS Code and a map application.

## Proposed solution

Add a "Show on Map" feature accessible from:

1. **Single cell** — right-click a cell containing geodata → "Show on Map"
2. **Selected range** — select multiple cells/rows → "Show on Map" (plots all points)
3. **Column context menu** — right-click column header → "Show Column on Map"

### Supported data formats

| Format | Example | Detection |
|---|---|---|
| PostGIS geometry/geography | WKB hex, WKT | Column type \`geometry\`, \`geography\` |
| GeoJSON | \`{"type":"Point","coordinates":[lng,lat]}\` | JSON with \`type\` + \`coordinates\` |
| JSON lat/lng | \`{"lat":51.5,"lng":-0.1}\` or \`{"latitude":...}\` | Key pattern matching |
| Array \`[lng, lat]\` | \`{-0.1,51.5}\` or \`[-0.1, 51.5]\` | 2-element numeric array |
| Separate lat/lng columns | \`lat\` + \`lng\` columns | User selects which columns |

### Map features

- **Webview panel** with interactive map (Leaflet.js or Mapbox GL JS)
- **Multiple points** from selected rows plotted simultaneously
- **Label column** — user can pick a column to display as point labels on the map (e.g. city name, ID)
- **Popup on click** — show full row data for each point
- **Cluster** large datasets (1000+ points)
- **Auto-zoom** to fit all points
- **Color by value** — optionally color points by a numeric/categorical column

### UI flow

1. Select cells or rows containing geodata
2. Right-click → "Show on Map"
3. If format is ambiguous, prompt: "Which columns contain coordinates?"
4. If label column is available, prompt: "Label points by?" (dropdown of text columns)
5. Map opens in a webview panel

## Alternatives considered

- External tools (QGIS, kepler.gl) — requires export, context switch
- VS Code map extensions — don't integrate with database results
- PostGIS viewer in pgAdmin — PG only, not in VS Code

## Implementation notes

- Use Leaflet.js (open source, ~40KB, no API key required) with OpenStreetMap tiles
- WKB parsing: use a small WKB decoder (or \`wkx\` npm package)
- GeoJSON: native JSON parsing
- Webview panel similar to result panel
- Limit: render up to 10,000 points (configurable), warn if more
- For PostGIS binary: detect by column type (\`geometry\`, \`geography\`) and parse hex WKB

Siyet:
## Summary
- Adds a 🗺 button to the Result Panel toolbar that opens a Leaflet-based map for the current result set.
- Auto-detects a wide range of coord formats: GeoJSON Point, WKT `POINT(lng lat)`, `{lat,lng}` / `{latitude,longitude}` objects, `[lng,lat]` arrays, PG brace-array strings (`{lng,lat}`), and separate `lat`/`lng` columns.
- Markers show a hover tooltip (configurable label column) and a popup with full row data; auto-zoom fits all points; hard cap of 10 000 markers per view.

Closes #7

## Assumptions
- **Binary WKB (hex) is not parsed.** The issue mentions WKB parsing via `wkx` as an implementation note. Instead of adding a native parser, the contract is that drivers return WKT or GeoJSON — PG can do this with `ST_AsText()` / `ST_AsGeoJSON()`. Surfacing hex WKB as unparseable is safer (no silently-wrong coordinates) and avoids pulling a new dependency into the core code path. Follow-up work can wire a `wkx` call inside the PostGIS path of `postgres.ts`.
- **Clustering and "color by value" are deferred.** Both are listed as features in the issue but require additional plugins (leaflet.markercluster) / chart-style config UI. The MVP focuses on the core "point + popup + zoom" flow; the extension points (mode/label selectors) are wired so these can be layered in without breaking message compatibility.
- **Column / cell context menus not added.** The issue mentions "right-click cell / column header → Show on Map". Only the Result Panel toolbar button is wired in this PR; the command accepts the same payload shape as `viewstor.visualizeResults`, so cell/column entry points are a one-liner to add later.
- **OpenStreetMap tile server with no API key.** Default `{s}.tile.openstreetmap.org`. CSP allows `https:` img-src so tiles render; switching to Mapbox would require a settings-backed API key and is out of scope.
- **Point limit 10 000 (hard-coded default).** Matches the issue's "up to 10 000 points (configurable)". The `MapPanelManager.show` signature accepts an override, but there is no user-facing setting yet — if needed that's a small addition to `package.json` > `configuration` (`viewstor.mapPointLimit`).
- **No dedicated "pick lat / lng columns" modal.** Auto-detection runs on open; the sidebar dropdowns let the user switch coord column or label column. If no format is detected at all, a warning message explains what's supported rather than forcing a multi-step picker.

## Manual test cases
1. **Golden path — PG single-column geometry.** In a PG connection, `SELECT id, ST_AsGeoJSON(geom)::json AS geom, name FROM cities LIMIT 100` → open results → click 🗺. Expect: map with 100 markers, hover tooltips show `name`, popup shows all columns, view fits all points.
2. **lat / lng pair.** `SELECT id, lat, lng, label FROM airports` → 🗺. Expect: auto-detects the `lat`/`lng` pair (not a single-column mode). Label dropdown defaults to `label`.
3. **WKT strings.** `SELECT 'POINT(-0.1 51.5)' AS g UNION ALL SELECT 'POINT(2.35 48.86)'` → 🗺. Expect: 2 markers (London, Paris), auto-zoom fits both.
4. **No coords detected.** Run a query with no geo data (e.g. `SELECT 1, 2`) → 🗺. Expect: warning toast "No geographic coordinates detected…"; no empty panel left open.
5. **Mixed garbage.** Rows where half have valid coords and half have `NULL` or out-of-range values (lat=999). Expect: status shows `N points · M skipped`; skipped rows don't crash the panel.
6. **Truncation.** Query with 15 000 valid points. Expect: 10 000 markers rendered, status bar shows `10000 points · truncated to 10000 of 15000`.
7. **Swap label column.** With a plotted result, change the "Label" dropdown from `name` to `id`. Expect: tooltips update in place, map zoom/center preserved.
8. **Theme compatibility.** Switch VS Code between light and dark themes while a map is open. Expect: popup background, borders, and toolbar all follow the theme (marker tiles stay OSM).

## Checklist
- [x] **CHANGELOG.md** — `[Unreleased]` entry with `(#7)` link
- [x] **CLAUDE.md** — new "Map View" section documenting the module layout and Leaflet bundling
- [x] **README.md** — feature matrix row + "Map View" section
- [x] **l10n** — `command.showOnMap` added to `l10n/nls/package.nls.json`
- [x] **Unit tests** — `mapDataTransform.test.ts` (28 tests covering GeoJSON / WKT / pair / PG-array / invalid input / label suggestion)
- [x] **extension.vscode.test.ts** — added `viewstor.showOnMap` to the required-commands array
- [x] **resultPanel.test.ts** — assert the `mapBtn` button is rendered
- [x] **Wiki** — a follow-up Wiki page "Map view" is recommended (formats, OSM attribution, 10k limit); not pushed from this PR


Siyet (commented):
Local gate: `tsc --noEmit` clean, `eslint` 0 errors / 18 pre-existing warnings, `npm run build` succeeds (only the known `pg-native` warning). `npm test` — 23/24 test files pass (`sqliteRebuild.test.ts` worker crashes before asserting; pre-existing, unrelated to this PR).

### Accuracy
- `src/views/resultPanel.ts:1527` posts `pageRows`, not the full result set. In table mode (server-side pagination) the user can only ever see up to `pageSize` markers (≤1000), not the documented "up to 10,000". The issue spec says "plot all points from selected rows" and the PR advertises a 10k cap — in practice the cap is never reachable from a table view. Either wire a "fetch all" path like export, or state this limitation in the README / warning toast.
- Issue lists cell / column-header / selection entry points; only the toolbar button is implemented. Called out in PR body as deferred, but the toolbar button treats "all visible rows" as the selection, which is different from the selection-aware flow the issue describes. OK to defer — just make sure the follow-up keeps the `viewstor.showOnMap` payload shape stable (currently `{ columns, rows, color?, tableName?, schema?, connectionName? }`).

### Bugs
- **XSS in tooltip** — `src/webview/scripts/map-panel.js:88`: `marker.bindTooltip(String(labelValue), …)`. Leaflet's default tooltip renders the content string as HTML (inserted via `innerHTML`). CSP has `script-src 'unsafe-inline'`, so a label cell with e.g. `""` executes. The popup path (`renderRowPopup`) already escapes — do the same here: `marker.bindTooltip(escapeHtml(String(labelValue)), …)`. The PR body omits XSS from the manual test cases; add one.
- **Over-eager auto-detect** — `src/map/mapDataTransform.ts:77–81` (the "1b" fallback) returns the first column where *any single row* parses. Combined with the permissive paths in `parseGeoValue`:
  - array path (line 184): any 2+ element numeric array becomes `{lat: arr[1], lng: arr[0]}`. An integer-array column like `tags int[]` with `{40,70}` gets picked as geo.
  - plain-pair path (line 238): any `"x, y"` string where both sides are finite numbers matches. A column like `"42, 7"` representing `price, discount` false-positives.
  
  Either drop the 1b fallback, or require ≥3 of the first N rows to parse (and still short-circuit those two loose paths behind a type/name hint).
- **Stale panel reuse** — `src/map/mapPanel.ts:61`: `panelKey = \`map:\${panelTitle}\``. Title in `mapCommands.ts:26` is built from `tableName` alone, so two tables with the same name in different schemas collide into one panel (the second call mutates the first panel's state). Include `schema` + `connectionId` in the key; keep the title human-readable separately.
- **Redundant `setTimeout(..., 100)`** — `mapPanel.ts:104`. The webview already posts `{type: 'ready'}` once its script runs, which triggers `sendData` via the handler. The 100 ms is a workaround, not a fix, and can fire before the webview is ready on slow machines (message gets dropped when the panel hasn't finished loading). Drop the timeout; rely on `ready`.
- **`retainContextWhenHidden` + unconditional `html` rewrite** — `mapPanel.ts:98` reassigns `webview.html` on every `show()` call, including when revealing an existing panel. That throws away the Leaflet state (zoom, pan) the user configured. Only set `html` on create; on reveal just re-post data.

### Performance
- `map-panel.js:91` calls `renderRowPopup(pt.row, msg.columns)` eagerly for every marker. At 10k markers × N columns that's a meaningful GC/string-concat cost before the user clicks anything. Leaflet accepts a function: `marker.bindPopup(() => renderRowPopup(pt.row, msg.columns))` — lazy by design.
- No clustering. Issue explicitly calls out "Cluster large datasets (1000+ points)". Acknowledged as deferred in PR body, but the status bar only warns at 10k truncation — with 5k uncluster-ed markers on a low-end machine Leaflet visibly stutters. At minimum, warn at `>1000` points that clustering is not yet supported.
- `getBounds()` on a `FeatureGroup` of 10k markers is O(N). Only called once per render, so acceptable, but combined with the above this compounds.

### Robustness
- `QueryColumn.dataType` isn't uniform across drivers (e.g., ClickHouse returns `Point`, PG returns `geometry` / `geography`, SQLite has no native geo type). The regexes in `SINGLE_TYPE_HINTS` (`/point/i` in particular) will also match unrelated types like `breakpoint_id`. Anchor them: `/^(geometry|geography|point|geojson)$/i`.
- `parseGeoValue` ignores the Z/M coordinate of WKT Points (line 203) — fine for mapping, but should be noted.
- No `activationEvents` update; relies on the existing `onView:viewstor.connections`. If a user opens the Command Palette before the tree view ever loads, `viewstor.showOnMap` activates the extension via the implicit `onCommand:` contribution — works, but worth confirming in the VS Code activation test.
- `resultPanel.ts:229-237` forwards the entire `rows` array through the webview → host IPC boundary on click. For a 1000-row page with wide rows, that's a non-trivial postMessage payload serialised twice. Consider keeping rows host-side and passing only `panelKey`.

### Tests
- Good unit coverage for pure logic. Missing regressions that would have caught the issues above:
  - `detectCoordMode` on `[{tags: [40, 70]}]` without name/type hint — should return `null`, currently returns `{kind:'single', column:'tags'}`.
  - `detectCoordMode` on `[{price_pair: '42, 7'}]` — should return `null`.
  - `parseGeoValue` on an array of length ≥3 (currently happily takes `[lng=arr[0], lat=arr[1]]`, discarding the rest — intended?).
  - `toFiniteNumber` on `BigInt` — covered implicitly, add explicit test.
- No test for the webview message round-trip (resultPanel button → command → MapPanelManager). `resultPanel.test.ts` only asserts the button's presence. At least assert that clicking it posts `{type:'showOnMap', columns, rows}`.
- `extension.vscode.test.ts` required-commands list updated — OK.
- No driver contract change — correct, no optional method added.

---

Siyet:
## Review

Thorough, well-scoped PR — map feature is cleanly isolated, `mapDataTransform.ts` is pure/tested (28 tests, all edge cases), Leaflet assets are wired into webpack correctly, CSP is tight (no `unsafe-eval`), and webview resources are vendored (no CDN pulls).

### Auto-fix pushed (f145e02)

- **Dead `last*` state in `src/webview/scripts/map-panel.js`** — `lastColumns`, `lastMode`, `lastLabel` were assigned in `renderPoints()` and `void`-suppressed at the bottom of the IIFE but never read. Removed.

### Findings left for author

1. **Permissive single-column fallback in `detectCoordMode` (`src/map/mapDataTransform.ts:77-81`).** When no column matches the type/name hints, the final fallback picks "any column where at least one sample parses as a coordinate." This will happily treat e.g. an `extra_info` JSONB column as the geo column if *one* row happens to contain a `{lat, lng}`-looking blob, producing surprising maps for non-geo tables. Consider either dropping this fallback entirely, or requiring a higher hit rate across sampled rows (e.g. ≥50%) before accepting.

2. **Panel key collision across connections (`src/map/mapPanel.ts:61`).** The panel key is `` `map:${panelTitle}` `` and titles are built as `Map — ${tableName}` in `mapCommands.ts:27`. Two connections with a table named `users` will share one panel, and opening the second will repurpose the first. Incorporate `connectionId` into the key (matching the pattern used elsewhere), or include it in the title.

3. **Redundant `setTimeout(100)` in `mapPanel.ts:104`.** The webview posts `{type: 'ready'}` on load, and the host already handles that. The extra `setTimeout` is belt-and-suspenders that can fire before the DOM script has hooked its listener on slow reloads. Safe to drop; the `ready` handshake is sufficient.

4. **Minor:** `Map — ${connectionName}` fallback when no `tableName` (same file, line 29) — fine as-is, just noting that custom-query results won't get a stable panel identity. Not worth changing unless (2) is addressed.

### CI

⚠️ **E2E Tests** check is failing on the commit. My pushed fix is a pure webview-JS dead-code removal and cannot affect Node.js E2E driver tests, so this appears to be pre-existing on the PR head. I was unable to access job logs to diagnose the root cause — please investigate and rerun or push a fix. Other checks (Lint, Build, Unit Tests, Security Audit, Changeset Size) are green.

Not adding `reviewed` label until CI is green.

---

Siyet (commented):
## Fixes pushed
- `c3047f3` — `mapPanel.ts`: dropped the `setTimeout(…, 100)` in `show()`. The webview's `ready` message already triggers `sendData`, and reassigning `webview.html` re-executes the script so a fresh `ready` arrives on reveal too. The timeout caused a duplicate send on first open and could post to an already-disposed panel if the tab was closed within 100 ms.
- `19a3dee` — `mapDataTransform.ts`: tightened `hasAnyParseableSingle` to require `isValidCoord(lat, lng)`. Without that, the loose 1b path in `detectCoordMode` catches any text column whose first cell is a numeric pair (e.g. `"500, 10"` in a non-geo column matches `parseGeoString`'s pair regex) and shadows a legitimate `lat`/`lng` pair on the same result set. Existing tests still green.
- `fe8d85f` — empty commit to retrigger CI after an unrelated flaky E2E failure on the prior run (same diff, same code paths; second run passed).

## Issues to address
- WKB deferral is acknowledged in the PR description, but a row whose `geom` column returns PostGIS hex (`0101000020…`) will silently count as `skipped`. Worth a follow-up to surface a one-time hint in the status bar (e.g. *"N rows appear to be binary WKB — cast to `ST_AsGeoJSON(geom)` in your query"*) so users know why their data didn't plot.
- `mapPanel.ts:61` — `panelKey = \`map:${panelTitle}\`` collides when two different sources produce the same title (e.g. two connections both showing `Map — users`). Minor, but reusing the result panel's `(connectionId, schema, tableName)` key would eliminate the ambiguity.
- `mapPanel.ts:166-170` — `iconPaths` is embedded as a single-quoted HTML attribute (`data-icons='${iconPaths}'`) but `JSON.stringify` doesn't escape single quotes. Webview URIs don't contain `'` today, but HTML-escaping the JSON (same trick as `safeJsonForScript` elsewhere) would make this robust against future URI scheme changes.


---

Issue #67, opened 2026-04-13
Data diff: database statistics comparison (row count, size, bloat)


Siyet:
## Context

When comparing the same table across environments (dev vs staging), it's useful to see statistics differences alongside row and schema diffs.

## Proposed solution

Add a **Statistics Diff** tab (or section in Schema Diff) that compares:

### Table-level statistics
| Metric | PG source | CH source | SQLite source |
|---|---|---|---|
| Row count | `pg_class.reltuples` | `system.tables.total_rows` | `COUNT(*)` |
| Table size | `pg_table_size()` | `system.tables.total_bytes` | `dbstat` (if available) |
| Index size | `pg_indexes_size()` | — | — |
| Total size | `pg_total_relation_size()` | `system.tables.total_bytes` | — |
| Dead tuples | `pg_stat_user_tables.n_dead_tup` | — | — |
| Last vacuum | `pg_stat_user_tables.last_vacuum` | — | — |
| Last analyze | `pg_stat_user_tables.last_analyze` | — | — |
| Sequential scans | `pg_stat_user_tables.seq_scan` | — | — |
| Index scans | `pg_stat_user_tables.idx_scan` | — | — |

### Display
- Side-by-side comparison with visual indicators (green = better, red = worse)
- Human-readable sizes (KB/MB/GB)
- Percentage difference for numeric values

### Conditions
- Only available when comparing tables from the same database type
- Only shown when the driver supports statistics (Redis excluded)
- PG has the richest stats; CH and SQLite show what's available

## Implementation notes

- Add `getTableStatistics?(name, schema)` optional method to `DatabaseDriver`
- Return a `Record<string, string | number>` for flexibility
- Add `computeStatsDiff()` to `diffEngine.ts`
- New section in diff panel webview

Issue #66, opened 2026-04-13
Data diff: schema diff should include indexes, sequences, triggers, constraints


Siyet:
## Context

The current Schema Diff tab (#5) compares only columns (name, type, nullability, PK status). This misses important structural differences between tables.

## Proposed additions

Extend Schema Diff to compare:

### Indexes
- Index name, columns (ordered), uniqueness, type (btree/hash/gin/gist), partial index predicate
- Source: PG `pg_indexes` + `pg_index`, CH `system.data_skipping_indices`, SQLite `PRAGMA index_list` + `PRAGMA index_info`
- Display: added/removed/changed indexes (e.g., same columns but different type)

### Constraints
- CHECK, UNIQUE, FOREIGN KEY constraints
- Source: PG `pg_constraint`, SQLite `PRAGMA foreign_key_list`
- Display: constraint name, type, columns, referenced table (FK)

### Sequences
- Sequence name, start value, increment, min/max
- PG-only (`pg_sequences`)

### Triggers
- Trigger name, event (INSERT/UPDATE/DELETE), timing (BEFORE/AFTER), function
- Source: PG `information_schema.triggers`, SQLite `sqlite_master`

## Implementation notes

- Add `getTableObjects(name, schema)` optional method to `DatabaseDriver` interface, or extend `getTableInfo()` return type
- Add `IndexInfo`, `ConstraintInfo`, `TriggerInfo` to `src/types/schema.ts`
- Extend `computeSchemaDiff()` in `diffEngine.ts` to compare these objects
- Add sections to Schema Diff tab in webview (accordion or sub-tables)
- Update `driverContract.test.ts` with new optional methods

## Not in scope

- Full DDL diff (use `showDDL` for that)
- Cross-database-type comparison (PG indexes vs CH data skipping indices have different semantics)

Issue #59, opened 2026-04-13
P0: Extension activation smoke test, connectionManager tests, split commands/index.ts


Siyet:
## Problem

Regression bugs keep appearing when fixing or adding features. Root cause analysis shows:

1. **Zero test coverage for core modules** — `connectionManager.ts` (429 lines, imported by 12 files), `commands/index.ts` (1598 lines, 40+ commands), `extension.ts` (388 lines) have no tests
2. **Extension activation failures recurred twice** (v0.2.2 NLS, v0.3.1 SQLite native) — CI doesn't validate activation
3. **God object** — `commands/index.ts` is 1598 lines with shared mutable state; any command change can break neighbors

### 1. Extension activation smoke test in CI
- Expand `extension.vscode.test.ts` to verify:
  - Extension activates without errors
  - All registered commands are available
  - Native modules load or gracefully fallback
- Ensure this runs in CI on every PR

### 2. Unit tests for connectionManager
- `loadConnections` / `saveConnections` round-trip
- `connect` / `disconnect` lifecycle
- `getDriverForDatabase` concurrent calls
- Merge user + project connections
- Folder CRUD with nesting

### 3. Split commands/index.ts into focused modules
```
src/commands/
  index.ts              — only registerCommands(), routing
  queryCommands.ts      — runQuery, cancelQuery, explainQuery
  tableCommands.ts      — showTableData, editCell, insertRow, deleteRows
  exportCommands.ts     — exportAllData, copyAs*
  connectionCommands.ts — addConnection, editConnection, removeConnection
  treeCommands.ts       — refresh, filter, fold/unfold
```
Each file owns its scope and state. Changes in tableCommands can't break queryCommands.

## Why P0

These are the highest-leverage changes: they cover the most-imported untested modules and eliminate the #1 regression source (monolithic commands file).

PR #79, opened 2026-04-15
fix(chart): re-enable Grafana export, fix JSON shape and SQLite SQL

### LINKED ISSUE #50: Grafana export: fix SQL dialect, datasource type, and import flow
## Problem

Grafana export generates incorrect SQL and datasource config for non-PostgreSQL connections:

- **SQL dialect**: uses `date_trunc()` for SQLite/ClickHouse instead of `strftime()` / `toStartOf*()`
- **Datasource type**: always `grafana-postgresql-datasource` regardless of actual DB type
- **Schema**: includes `"public"` schema for SQLite (which has no schemas)
- **Import**: Grafana import page shows empty Name field despite `title` being set in JSON

## Current state

- Button hidden (`display:none`) until fixed
- Partial fix in `chartPanel.ts` — SQL is now rebuilt with correct `databaseType` via `buildAggregationQuery`
- `DB_TYPE_TO_GRAFANA_DS` mapping exists in `grafanaExport.ts` (sqlite → `frser-sqlite-datasource`)

## TODO

- [ ] End-to-end test: export → import into Grafana for each supported DB type
- [ ] Fix JSON structure for Grafana import API (`/api/dashboards/import` vs `/api/dashboards/db`)
- [ ] Handle `pushToGrafana` correctly for each datasource
- [ ] Add dashboard name input before export
- [ ] Re-enable button after fixes

Related: #31

Siyet:
## Summary
- Flatten `buildGrafanaDashboard()` output so Grafana's UI import reads `title` correctly; added `__inputs` / `__requires` placeholders so the import wizard picks the right datasource plugin.
- New `buildApiPayload()` helper + updated `pushToGrafana()` wrap the flat dashboard in the `/api/dashboards/db` envelope and strip UI-only fields.
- Prompt for dashboard name before export; remove `display:none` on the Export-to-Grafana button (visibility is now driven by chart-type compatibility in the webview).
- `buildAggregationQuery` / `buildFullDataQuery` drop the schema prefix for SQLite (SQLite has no schemas; stored `"public"` was producing unparseable SQL).

Closes #50

## Assumptions
- **Flat dashboard JSON is the canonical export shape.** The issue calls out "Import: Grafana import page shows empty Name field despite `title` being set in JSON", which is a symptom of exporting the API-envelope shape (`{dashboard:{title:…}}`) while the Grafana UI import reads the top-level `title`. The fix is to return the flat shape everywhere the user touches JSON (copy / save to file), and wrap it only server-side in `pushToGrafana`. This matches the convention used by dashboards exported from Grafana itself.
- **`__inputs` placeholder only when no concrete UID is bound.** When `databaseType` is known (postgresql / clickhouse / sqlite) but the caller doesn't pass a datasource UID, we emit `__inputs` so the import wizard shows a picker. When a UID *is* passed, we assume the caller already knows which datasource to wire up and skip placeholders.
- **`pushToGrafana` only stripping `__inputs`/`__requires`.** The API rejects UI-only fields; we keep everything else (tags, time range, schemaVersion) untouched so behaviour stays close to Grafana's own round-trip.
- **Panel datasource mirrors target datasource.** Older Grafana versions required `datasource` at the panel level too; setting both is harmless for current versions and removes one class of "unassigned datasource" UI glitches on import.
- **No live Grafana end-to-end test.** The "End-to-end test: export → import into Grafana for each supported DB type" acceptance item requires a running Grafana; that is deferred. Verification in this PR is the updated unit tests plus manual smoke steps below.

## Manual test cases
1. **PostgreSQL → Grafana UI import (golden path).** Run a PG query → open chart → pick Line → click Export to Grafana. Enter a dashboard name in the prompt. Save the JSON. In Grafana → Dashboards → New → Import → Upload JSON. Expect: Name field pre-filled from prompt, datasource picker shows PostgreSQL option, import succeeds, panel renders with the query.
2. **SQLite → export.** Open an SQLite connection, show table data, open chart. Confirm the Export to Grafana button is enabled for line/bar/etc (disabled for radar/treemap/etc). Export. Expect: generated `rawSql` has no `"public"."…"` prefix, `__inputs[0].pluginId === 'frser-sqlite-datasource'`.
3. **ClickHouse → export.** Same as above but for a CH connection. Expect: `rawSql` uses `toStartOf…()` (not `date_trunc`), datasource type is `grafana-clickhouse-datasource`.
4. **Cancel the name prompt.** Click Export to Grafana → press Escape on the input box. Expect: no JSON popup opens, no error notifications.
5. **Empty dashboard name.** Clear the input and try to confirm. Expect: inline "Name cannot be empty" validation, confirm disabled.
6. **Push to Grafana (requires `viewstor.grafanaUrl` + `viewstor.grafanaApiKey`).** After export, click Push. Expect: request body has the `{dashboard, overwrite, message}` envelope (verify in Grafana server logs / reverse proxy), success toast with dashboard URL.
7. **Incompatible chart type.** Switch chart to Radar. Expect: Export button becomes disabled with tooltip "This chart type has no Grafana equivalent".
8. **Unknown database type.** In-memory ad-hoc chart with `databaseType` unset. Expect: dashboard builds, but no `datasource` / `__inputs` block — user manually picks on import.

## Checklist
- [x] **CHANGELOG.md** — `[Unreleased] / Fixed` entry with `(#50)` link
- [x] **CLAUDE.md** — `grafanaExport.ts` line updated to describe the flat shape + `buildApiPayload`
- [x] **Unit tests** — `grafanaExport.test.ts` rewritten for the flat shape, adds `__inputs`/`__requires` cases, adds `buildApiPayload` suite; `chartTypes.test.ts` gets a SQLite-schema-stripping suite
- [ ] **README.md** — not touched; Grafana export isn't a top-level README feature yet (the button was hidden in prior releases)
- [ ] **l10n** — new strings ("Grafana dashboard name", "Used as the dashboard title in Grafana.", "Name cannot be empty") are fed through `vscode.l10n.t(...)` and will be picked up by the standard extract step; no manual `package.nls.json` changes required
- [ ] **Wiki** — follow-up Wiki page "Grafana export" once the end-to-end test is wired up; not pushed from this PR


Siyet (commented):
Local checks: `npx tsc --noEmit` clean, `npm run lint` 18 pre-existing warnings, `npm test` 767 passed (one unrelated tinypool worker-exit warning), `npm run build` succeeds (pg-native warning is expected).

### Accuracy
- `CHANGELOG.md:10` — inserting `### Fixed` between the existing Added entry and its siblings re-parents the three items that used to live under Added (Schema diff indexes/constraints #66, Data diff statistics tab #67, Regression prevention #59/#60/#61) into Fixed. They are features, not fixes. Add the Fixed block after the Added list, or re-add an `### Added` header above "Schema diff:".
- Issue #50 acceptance item "End-to-end test: export → import into Grafana for each supported DB type" is explicitly deferred (noted in the PR body). Fine to track as a follow-up, but `Closes #50` will auto-close an issue whose acceptance isn't fully met — consider `Refs #50` until the E2E is wired up.

### Bugs
- `src/chart/grafanaExport.ts:105-115,130` + `src/chart/chartPanel.ts:254,300` — the Push-to-Grafana path is still broken for the common case. `buildGrafanaDashboard(config)` is always called without a `datasourceUid`, so every panel/target ends up with `uid: "${DS_DEFAULT}"`. `buildApiPayload` strips `__inputs`/`__requires` but leaves `${DS_DEFAULT}` on `panel.datasource.uid` and `target.datasource.uid`. The `/api/dashboards/db` endpoint does not resolve input variables — it will either 400 or store a dashboard pointing at a non-existent datasource. This matches the "Handle `pushToGrafana` correctly for each datasource" TODO in #50, which the PR description/CHANGELOG imply is fixed ("`pushToGrafana` wraps the payload …"). It isn't. Options: resolve `${DS_DEFAULT}` to a real UID at push time (lookup via `/api/datasources` by `type`), fail-fast with a clear message when the dashboard still contains `${DS_...}` placeholders, or drop the Push button until the resolution path is implemented.
- `src/chart/grafanaExport.ts:173-174` — the destructuring comment says "force id/uid to be absent so Grafana assigns new ones (prevents 'Dashboard not found' errors when re-pushing)" but the implementation only strips `__inputs` / `__requires`; `id`/`uid` are preserved if they exist on the input object. Either strip them (`const { __inputs, __requires, id, uid, ...rest } = dashboard as …`) or fix the comment.
- `src/chart/grafanaExport.ts:249` — on a push where Grafana returns no `url` in the response body, the function returns the bare Grafana base URL instead of something useful. Minor, but the success toast then shows the server root rather than a dashboard link.
- `src/chart/chartPanel.ts:219-222` — `defaultName` falls back to `state.panel.title` which is `'Chart'` (hard-coded in `show()` via `panelTitle = title || 'Chart'`). For ad-hoc queries with no `tableName` and no `msg.config.title` the prompt preloads "Chart", which becomes the Grafana dashboard title. Fine, but `'Viewstor Export'` would match the documented default on line 120 / 133. Low impact.

### Performance
OK.

### Robustness
- `src/chart/grafanaExport.ts:252-256` — `isApiPayload` only checks for a nested `.dashboard.panels` field. A user who hand-edits the exported flat JSON and accidentally wraps it in `{ dashboard: { … } }` before pushing will be detected as an API payload and the overwrite/message options from `pushToGrafana(opts)` get silently dropped. Not a P0; document the precondition in the JSDoc or accept both shapes and always merge `opts`.
- `src/chart/chartPanel.ts:297` — `JSON.parse(json) as GrafanaDashboard` with no validation. If the stored `json` was tampered with via `showGrafanaJson` flow it still gets trusted. Consider a minimal shape check before push.
- `l10n/bundle.l10n.*.json` — new strings `"Grafana dashboard name"`, `"Used as the dashboard title in Grafana."`, `"Name cannot be empty"` are wrapped in `vscode.l10n.t(...)` but none of the existing translation bundles contain them. PR body says the extract step will pick them up; confirm the build/publish pipeline actually runs that extract before release, otherwise these stay English-only for all locales.

### Tests
- No unit test for `pushToGrafana`. The whole point of this PR is fixing the push/import flow, but only `buildGrafanaDashboard` + `buildApiPayload` are exercised. A `vi.spyOn(globalThis, 'fetch')` test to assert (a) the URL is `/api/dashboards/db`, (b) the body is a `{dashboard, overwrite: true}` envelope, and (c) `__inputs`/`__requires` are absent from the posted body — would have caught the `${DS_DEFAULT}` push bug above.
- `src/test/grafanaExport.test.ts` — no coverage for the `sourceQuery` dialect (PG `date_trunc` vs CH `toStartOf*` vs SQLite `strftime`) going through to `target.rawSql`. `chartTypes.test.ts` tests the query builder in isolation; a single integration test in `grafanaExport.test.ts` covering `buildGrafanaDashboard({databaseType: 'sqlite', sourceQuery: '…'})` would lock in the #50 acceptance criterion.
- Flat-shape migration: no test confirms that consumers which used to read `result.dashboard.panels` (MCP `exportGrafana` in `src/mcp/server.ts`, standalone MCP `export_grafana_dashboard` in `src/mcp-server/index.ts`) were updated to the new shape. Please grep for `.dashboard.` accesses and add a regression test for any MCP consumer.

---

PR #81, opened 2026-04-15
feat: pgvector support in PostgreSQL driver

### LINKED ISSUE #14: feat: add pgvector support (PostgreSQL extension)
## Problem

pgvector is the most popular vector search extension for PostgreSQL. AI/ML engineers need to inspect vector embeddings, run similarity searches, and manage vector indexes — currently falling back to raw SQL.

## Proposed solution

Extend the existing PostgreSQL driver with pgvector-aware features. This is NOT a new driver — it enhances the PG driver.

  - Detect pgvector: `SELECT * FROM pg_extension WHERE extname = 'vector'`
  - Schema view: show `vector` columns with dimension in detail (e.g., `vector(1536)`)
  - Result grid: display vectors as truncated arrays with dimension count badge (e.g., `[0.012, -0.034, ...] (1536d)`)
  - getTableInfo: parse `vector(N)` type, show dimension
  - DDL: include `CREATE INDEX ... USING ivfflat/hnsw` syntax
  - SQL autocomplete: add pgvector operators (`<->`, `<#>`, `<=>`) and functions (`l2_distance`, `cosine_distance`, `inner_product`)
  - Index hints: detect missing vector indexes on columns used with distance operators
  - Vector-specific context menu: "Find similar" → generate `ORDER BY embedding <-> '[...]' LIMIT 10` template
- **No new library needed** — pgvector uses standard `pg` protocol
- **Tests:** e2e with `ankane/pgvector` Docker image

## Alternatives considered

- Separate pgvector driver — unnecessary, it's a PG extension using the same protocol
- Dedicated vector DB UI — pgvector users want it integrated with their PG workflow

Siyet:
## Summary
- Detect pgvector via `pg_extension`; lazily cached per connection and cleared on `disconnect()`.
- Schema tree and `getTableInfo` render `vector(N)` using `pg_attribute.atttypmod` instead of bare `vector` (which is what `information_schema.columns.udt_name` returns for user-defined types).
- SQL autocomplete gains pgvector distance operators (`<->`, `<#>`, `<=>`, `<+>`) and helper functions (`l2_distance`, `l1_distance`, `cosine_distance`, `inner_product`, `vector_dims`, `vector_norm`, `l2_normalize`) on pgvector-enabled connections. No new runtime dependency — pgvector uses the standard `pg` wire protocol.

Closes #14

## Assumptions
- **Scope narrowed to detection, schema display, and autocomplete.** The issue lists additional items (`CREATE INDEX ... USING ivfflat/hnsw` in DDL, result-grid truncation of long vectors, index hints for missing vector indexes, a "Find similar" context menu). Those are independent follow-ups; this PR implements the core plumbing (`detectPgvector` + `formatVectorType`) that later work can layer on without touching the driver interface again. Vector values already come back from `pg` as string literals like `[0.1,0.2,...]`, so they render in the result grid today — truncation is a display polish, not a correctness gap.
- **`atttypmod` holds the dimension directly for pgvector.** pgvector's `vector_typmod_in` stores the parsed `(N)` as-is (unlike `varchar`, which offsets by 4). `-1` means "no dimension specified at column definition time"; we fall back to bare `vector` in that case.
- **Operators exposed as keyword-kind.** `<->` / `<#>` / `<=>` / `<+>` don't naturally trigger VS Code completion (they're symbol sequences, not word-like), but surfacing them as completion items still lets users pull them up via the completion popup and shows the descriptive detail. Functions use `function` kind so VS Code tags them with the function icon.
- **Detection is lazy, one-shot per connection.** `pg_extension` is checked only when the first schema/completion query fires, so non-pgvector databases pay nothing. The flag resets on `disconnect()` so reconnecting to a different server re-evaluates.
- **No l10n changes.** The new strings are driver-internal detail text rendered by existing tree/completion UI that already consumes plain strings; no `vscode.l10n.t(...)` call sites added.

## Manual test cases
1. **Golden path — column detail shows dimension.** Connect to a PG instance with `CREATE EXTENSION vector`. Create a table `CREATE TABLE docs (id int, embedding vector(1536))`. Expand the connection → schema → `docs` → columns. Expected: the `embedding` column shows `vector(1536)` in the detail row, not `vector`.
2. **Column info panel.** Right-click `docs` → Show Column Info (or open table data → inspect column metadata). Expected: `dataType === 'vector(1536)'` for the embedding column.
3. **Autocomplete — functions.** In a SQL editor bound to a pgvector-enabled PG connection, type `SELECT l2_` after placing the cursor in expression context. Expected: `l2_distance`, `l2_normalize` appear in the completion list with the "pgvector" detail annotation. Also verify `cosine_distance`, `inner_product`, `vector_dims`, `vector_norm`, `l1_distance` appear.
4. **Autocomplete — operators.** Trigger completion manually (Ctrl+Space). Expected: `<->`, `<#>`, `<=>`, `<+>` appear with their distance descriptions.
5. **No pgvector, no leakage.** Against a PG connection without the extension installed, open a SQL editor and trigger completion. Expected: none of the pgvector operators/functions appear in the list.
6. **Unspecified dimension.** Create `CREATE TABLE t (v vector)` (no dimension). Expected: schema tree shows `vector` (bare), not `vector(-1)` or `vector(0)`.
7. **Disconnect / reconnect.** Disconnect from a pgvector-enabled DB, switch to a non-pgvector DB under the same connection entry (e.g. via "additional databases"), trigger completion. Expected: pgvector items no longer appear; `detectPgvector` re-runs against the new database.

## Checklist
- [x] **CHANGELOG.md** — `[Unreleased]` entry with `(#14)` link
- [x] **CLAUDE.md** — new `src/drivers/pgvector.ts` module documented under the driver section
- [x] **Unit tests** — `src/test/pgvector.test.ts` covers `formatVectorType` and `pgVectorCompletionItems`
- [ ] **README.md** — not touched; README feature matrix doesn't currently mention PG extensions, and pgvector support is an extension-conditional enhancement to an existing driver
- [ ] **l10n** — no new UI strings
- [ ] **Wiki** — a follow-up "pgvector" page once the remaining items from #14 (DDL `USING ivfflat/hnsw`, index hints, result-grid truncation, Find-similar) land — not pushed from this PR


Siyet (commented):
## Fixes pushed
- `ef517a8` — `completionProvider.provideCompletionItems` was dropping every driver `CompletionItem` whose `kind` was `function` or `keyword`, so the new pgvector operators (`<->`, `<#>`, `<=>`, `<+>`) and helpers (`l2_distance`, `cosine_distance`, `inner_product`, `vector_dims`, `vector_norm`, `l2_norm­alize`, `l1_distance`) never reached VS Code. The general-context path now forwards those kinds via `toVscodeItem`, unblocking manual test cases #3 and #4.


---

Issue #71, opened 2026-04-14
Restrict LLM/MCP access per connection (hide from agents, encrypted credentials)


Siyet:
## Context

Viewstor exposes connections to LLM agents through two MCP surfaces:
- In-process MCP commands (\`src/mcp/server.ts\`): \`viewstor.mcp.listConnections\`, \`getSchema\`, \`executeQuery\`, \`getTableData\`, \`getTableInfo\`, \`visualize\`, \`exportGrafana\`, \`openQuery\`, \`openTableData\`
- Standalone stdio MCP server (\`src/mcp-server/index.ts\`): \`list_connections\`, \`get_schema\`, \`execute_query\`, \`get_table_data\`, \`get_table_info\`, \`add_connection\`, \`reload_connections\`, \`build_chart\`, \`export_grafana_dashboard\`

Copilot Chat (\`@viewstor\`) also resolves the active connection and forwards schema context to the model.

Today **every configured connection is visible and usable by the agent**, including production databases. There is no way to:
1. Exclude a specific connection from agent tools altogether (agent shouldn't even see it exists).
2. Allow the agent to \"see\" a connection for schema context but refuse \`execute_query\` / \`get_table_data\` against it.
3. Prevent an agent from reading raw credentials out of the stored config and opening its own session outside Viewstor (the standalone MCP server reads \`~/.viewstor/connections.json\` and \`.vscode/viewstor.json\` directly, and passwords currently live in globalState in plaintext).

For users working with production or customer-data connections this is a real blocker.

### 1. Per-connection agent access mode

Add a new field to \`ConnectionConfig\` (and \`ConnectionFolder\` with inheritance, like \`readonly\` / \`color\`):

\`\`\`ts
agentAccess: 'full' | 'schema-only' | 'none'
\`\`\`

- \`full\` (default for backwards compat, or \`schema-only\` if we want secure-by-default): agent can list, read schema, execute queries.
- \`schema-only\`: agent sees the connection and can call \`getSchema\` / \`getTableInfo\`, but \`executeQuery\` / \`getTableData\` return \`{ error: 'Agent access denied for this connection' }\`.
- \`none\`: connection is filtered out of \`listConnections\`, and all other tools return \"connection not found\" for its id — from the agent's perspective it simply does not exist.

Enforce in both \`src/mcp/server.ts\` and \`src/mcp-server/index.ts\` at a single gate (helper in connectionStore / ConnectionManager).

[...]

PR #85, opened 2026-04-15
feat(mcp): agent write approval + query classifier

### LINKED ISSUE #75: Agent MCP queries: EXPLAIN-based safe mode + human confirmation for destructive ops
## Context

Right now, when an LLM agent runs a query through Viewstor MCP, the only safety check is the `readonly` flag: non-SELECT on a readonly connection → error; everything else just runs. See `src/mcp/server.ts:56-62` (in-process) and the analogous `execute_query` in `src/mcp-server/index.ts`.

Two gaps:

1. **No EXPLAIN validation.** Human users get `viewstor.safeMode` (block/warn/off) which runs `EXPLAIN` before SELECTs and catches accidental full-table scans — see `src/commands/index.ts` and the "Safe Mode" section in `CLAUDE.md`. Agents skip all of that. An agent writing `SELECT * FROM orders WHERE lower(email) LIKE '%foo%'` against a 200M-row prod table will happily hammer the DB with a Seq Scan.

2. **No human-in-the-loop for destructive queries.** On a non-readonly connection, an agent can `UPDATE` / `DELETE` / `TRUNCATE` / `DROP` / `ALTER` without asking. Even on a `dev` profile that's nervous; on anything touching real data it's unacceptable.

Related but distinct issues:
- #71 — *whether* the agent can reach a connection (full/schema-only/none)
- #72 — anonymizing *returned* data
- This issue — *what the agent is allowed to run and under what oversight*

### 1. Reuse existing Safe Mode for SELECT-shaped queries

MCP executeQuery (both surfaces) runs the same `EXPLAIN` check the human pipeline already uses. Per-connection / global `viewstor.safeMode` applies:

- `block` → return `{ error, explain, kind: 'seq_scan_blocked' }`. Agent sees a structured error and can adapt (add WHERE, add LIMIT, give up).
- `warn` → run the query, but include a `warnings: [{ kind: 'seq_scan', explain }]` field in the response. Agent can choose to retry smarter on the next turn.
- `off` → no check.

This also means we already have driver-specific EXPLAIN support (PG EXPLAIN + Seq Scan detection, SQLite EXPLAIN QUERY PLAN + SCAN TABLE, CH EXPLAIN + Full) — we just lift it out of the command handler into a helper both the UI and MCP paths call.

Same auto-LIMIT behavior for SELECTs without LIMIT — apply before the MCP returns rows.

### 2. Destructive-query classifier

Pure function in `src/mcp/queryRisk.ts`:

```ts
export type QueryRisk =
  | { kind: 'read' }                      // SELECT/EXPLAIN/SHOW/WITH
  | { kind: 'write'; verb: 'INSERT'|'UPDATE'|'DELETE'|'MERGE'|'UPSERT' }
  | { kind: 'ddl'; verb: 'CREATE'|'ALTER'|'DROP'|'TRUNCATE'|'RENAME' }
  | { kind: 'admin'; verb: 'GRANT'|'REVOKE'|'VACUUM'|'ANALYZE'|'REINDEX' }
  | { kind: 'unknown' }

export function classifyQuery(sql: string): QueryRisk
```

Parses the leading keyword after comments/whitespace. Handles multi-statement scripts by returning the most-severe kind. Unit-tested (no driver needed).

`unknown` is treated as the most dangerous category — if we can't classify it, the human confirms.

### 3. Per-connection `agentWriteApproval` setting

New field on `ConnectionConfig` (bubbles up through folder inheritance, tied to the `profile` preset from #71):

```ts
agentWriteApproval: 'always' | 'ddl-and-admin' | 'never'
```

- `always` (safest, default for `prod`) — every write / DDL / admin query triggers a human-confirmation prompt.
- `ddl-and-admin` (default for `dev`) — writes (INSERT/UPDATE/DELETE) auto-approve; DDL and admin still prompt.
- `never` (explicit opt-out, loud UI warning) — no prompts. Reserved for ephemeral sandboxes.

`readonly` from today stays as a separate kill-switch — when set, no write is ever allowed regardless of this setting.

### 4. Confirmation flow

When an agent calls `executeQuery` with a query that needs approval:

**In-process (VS Code MCP, `src/mcp/server.ts`)**
- Show a VS Code modal with: connection name, the full SQL, classification ("DELETE — affects rows"), and the EXPLAIN output when available.
- Buttons: **Run · Run & remember for this session · Cancel**.
- "Remember for this session" caches approval by `(connectionId, classifyQuery(sql).kind)` for N minutes (configurable; default 5). Does NOT cache by exact SQL text — we don't want an agent to pre-approve one query and then piggy-back a different destructive one.
- Cancel → MCP returns `{ error: 'User denied execution', kind: 'user_denied' }`.

**Standalone MCP server (`src/mcp-server/index.ts`, separate Node process)**
- Can't pop a VS Code modal. Two options, configurable:
  1. **Strict:** refuse all writes by default (`agentWriteApproval` default of `always` effectively means "never" for the standalone process). Document that write/DDL via the standalone server is only possible by explicitly setting approval to `never` — i.e. user accepts the risk for that specific stdio setup.
  2. **Bridge:** optional companion channel (local HTTP listener / named pipe to a running VS Code instance) that forwards the confirmation to VS Code if it's open. Out of scope for v1 — document the limitation.

v1 goes with Strict. Document loudly.

### 5. Response shape additions

Agent-facing result of `executeQuery` gains optional fields:

```json
{
  "rows": [...],
  "warnings": [
    { "kind": "seq_scan", "message": "...", "explain": "Seq Scan on orders ..." }
  ],
  "classification": { "kind": "write", "verb": "UPDATE" },
  "autoLimitApplied": 1000
}
```

When blocked:
```json
{
  "error": "Sequential scan detected on 'orders' (205M rows). Relax viewstor.safeMode or add a WHERE clause on an indexed column.",
  "kind": "seq_scan_blocked",
  "explain": "..."
}
```

Structured errors > opaque strings so the agent can reason about the failure.

### 6. Audit log

Every agent query that ran goes to an output channel with:

```
2026-04-15 10:23:11 [mcp] conn=prod-pg classify=write:UPDATE rows=0 approved=user-clicked explain=index
```

So the user can retroactively check what the agent touched. This is also the place to surface "agent tried to DROP a table at 3am" in case of post-incident review.

### 7. UI indicators

- Connection form: new **Agent write approval** dropdown next to the existing Safe Mode row (and the anonymization toggle from #72, and the access mode from #71).
- Tree view: a small shield icon when `agentWriteApproval === 'always'` — the user sees at a glance which connections will prompt for every agent write.

### 8. Profile preset wiring (from #71 comment)

The profile table gets two new rows:

| Setting | `dev` | `prod` |
|---|---|---|
| `safeMode` for agent queries | `warn` | `block` |
| `agentWriteApproval` | `ddl-and-admin` | `always` |

## Out of scope

- Parsing complex multi-statement scripts perfectly. We take a pessimistic view (worst-kind wins) which is fine for a safety gate.
- Catching semantic footguns (e.g. UPDATE without WHERE) before the driver runs them — the classifier sees `UPDATE`, triggers approval, user reads the SQL in the dialog, that is the checkpoint.
- Budget / row-count caps ("refuse to run any agent query returning >1M rows"). Could be a follow-up tied to the auto-LIMIT.
- Full query parser — leading keyword is enough for classification. When in doubt, classify as `unknown` and confirm.

## Acceptance criteria

- [ ] `classifyQuery` in `src/mcp/queryRisk.ts` — pure, unit-tested, handles SELECT/INSERT/UPDATE/DELETE/TRUNCATE/DROP/ALTER/CREATE/RENAME/GRANT/REVOKE/VACUUM/ANALYZE/WITH; multi-statement returns most severe; `unknown` for unparseable input
- [ ] Existing Safe Mode EXPLAIN logic extracted from `commands/index.ts` into a reusable helper (`src/drivers/safeMode.ts` or similar)
- [ ] In-process MCP `executeQuery` runs Safe Mode check + classifier before dispatch; uses a VS Code modal for confirmation with SQL + EXPLAIN preview + "Remember for session" option
- [ ] Standalone MCP `execute_query`: writes refused by default when `agentWriteApproval !== 'never'`; documented clearly
- [ ] `ConnectionConfig.agentWriteApproval` persisted and editable via connection form
- [ ] Folder-level inheritance
- [ ] `profile: prod` (from #71) sets `agentWriteApproval: 'always'`; `profile: dev` sets `'ddl-and-admin'`
- [ ] Structured response: `warnings`, `classification`, `autoLimitApplied`, `kind` on errors
- [ ] Session approval cache keyed by `(connectionId, classification.kind)`; NOT by exact SQL; TTL configurable via `viewstor.agentWriteApproval.cacheSeconds` (default 300)
- [ ] Audit log to an output channel, one line per agent query
- [ ] Tree view shield icon on connections with `agentWriteApproval === 'always'`
- [ ] Unit tests for classifier (parametric table: input → expected kind/verb)
- [ ] Integration test: MCP executes SELECT with Seq Scan under `block` → returns structured error; UPDATE under `always` + user denies → returns `user_denied`
- [ ] README + wiki page: "Agent safety: EXPLAIN + confirmation", including threat model and what each `agentWriteApproval` mode actually blocks

Siyet:
## Summary
- Pure `classifyQuery(sql)` in `src/mcp/queryRisk.ts` that buckets SQL into `read` / `write` / `ddl` / `admin` / `unknown` (with verb). Multi-statement → most severe; unparseable → `unknown` (treated as destructive).
- `ConnectionConfig.agentWriteApproval` (`always` / `ddl-and-admin` / `never`) with `ConnectionFolder` inheritance via new `ConnectionManager.getAgentWriteApproval(id)`. Default `always`.
- In-process MCP `executeQuery`: VS Code modal with SQL preview + classification, `Run` / `Run & remember for session`. Session approval cached per `(connectionId, kind)` for 5 min. Denial → structured `{ kind: 'user_denied', classification }`.
- Standalone stdio MCP `execute_query`: cannot prompt, so refuses non-reads with `{ kind: 'approval_required', classification, approvalMode }` unless the connection is explicitly `agentWriteApproval: 'never'`.
- Connection form: new "Agent write approval" dropdown next to Safe Mode.

Closes #75

## Assumptions
- **Scope deliberately narrowed to the approval gate.** The issue lists additional items (EXPLAIN-based Safe Mode for SELECTs via MCP, auto-LIMIT at the MCP boundary, audit-log output channel, tree view shield icon, `profile: 'prod' | 'dev'` preset wiring). Those are each independent and none of them depend on code this PR touches. Shipping the classifier + approval gate first gives the hardest safety guarantee (no silent writes by an agent) and leaves the rest as incremental follow-ups. Documented below.
- **Default approval policy is `always`**, not `ddl-and-admin` — safer-by-default. Pre-existing connections with `agentWriteApproval: undefined` resolve to `always` via `getAgentWriteApproval()`, which means agents that currently run INSERT/UPDATE/DELETE via MCP will start prompting. That is the intended behavior and the entire point of the issue; users who relied on the old pass-through can set `never` per connection.
- **Session approval cache is in-memory only**, keyed by `${connectionId}:${risk.kind}` with a 5-min TTL. Not by exact SQL text (per issue spec: "we don't want an agent to pre-approve one query and then piggy-back a different destructive one"). TTL is a constant in `src/mcp/server.ts` for now; promoting to `viewstor.agentWriteApproval.cacheSeconds` is a one-line follow-up if needed.
- **Standalone MCP picks the strict path** described in the issue. Bridge (local HTTP to a running VS Code) is out of scope; documented in the error message so agents reading it can explain the situation to the user.
- **Folder inheritance** uses the same walk-up pattern as `readonly`, but follows `parentFolderId` (nested folders), returning the first non-undefined `agentWriteApproval`. Default `'always'`. A cycle guard (`seen` set) prevents pathological config from looping.
- **Readonly is still an unconditional kill-switch.** A readonly connection refuses non-read queries regardless of the approval mode, returning `{ kind: 'readonly_blocked' }` — that layering matches the existing behavior and keeps approval orthogonal.
- **No package.json settings keys added.** The acceptance-criteria list mentions `viewstor.agentWriteApproval.cacheSeconds`. Not exposing it yet keeps the surface small; can be added later without breaking anyone.

## Follow-ups (not in this PR, tracked under #75)
- EXPLAIN-based Safe Mode reused for MCP SELECTs (structured `{ kind: 'seq_scan_blocked' }` / `warnings[]`). Requires lifting the existing command-side Safe Mode into `src/drivers/safeMode.ts`.
- Auto-LIMIT for SELECTs at the MCP boundary, with `autoLimitApplied` in the response.
- Output-channel audit log (`2026-... [mcp] conn=... classify=... rows=... approved=...`).
- Tree view shield icon on `agentWriteApproval === 'always'`.
- `profile: 'prod' | 'dev'` preset from #71 that wires `agentWriteApproval` + `safeMode` + `readonly` in one switch.
- Configurable cache TTL setting.

## Manual test cases
1. **Golden path — modal prompt for DELETE.** With a connection whose `agentWriteApproval` is default (unset → `always`), call `viewstor.mcp.executeQuery` with `DELETE FROM users WHERE id = 1`. Expect: VS Code modal pops up with "Run" / "Run & remember for session" / "Cancel". Pick "Run". Expect: row deleted, response includes `classification: { kind: 'write', verb: 'DELETE' }`.
2. **Remember for session.** Same setup, pick "Run & remember for session". Expect: deletion runs. Immediately call another `DELETE FROM users WHERE id = 2` — no modal this time, runs through. Wait 6 min (or change TTL), call a third `DELETE`: prompt reappears.
3. **User denies.** Call an `UPDATE`, pick Cancel on the modal. Expect: response is `{ error: 'User denied execution', kind: 'user_denied', classification: { kind: 'write', verb: 'UPDATE' } }`. No rows changed.
4. **ddl-and-admin auto-runs writes.** Edit the connection, set "Agent write approval" to "DDL & admin only". Run an `INSERT` via MCP. Expect: runs without a prompt. Then run `DROP TABLE temp_agent_test`: prompt appears (DDL still gated).
5. **never passes everything through.** Set the connection to "Never". `INSERT`, `UPDATE`, `DROP`, `GRANT` — none of them prompt.
6. **Folder inheritance.** Create a folder with `agentWriteApproval: 'never'` (via `.vscode/viewstor.json` for project-scope or globalState edit) and put a connection inside that has no own setting. Call `executeQuery` with `DELETE`: no prompt, runs through. Override the connection's own setting to `'always'`: prompt reappears.
7. **Multi-statement severity.** Call `executeQuery` with `SELECT 1; DROP TABLE temp`. Expect: modal treats it as DDL (the most severe), classification in response is `{ kind: 'ddl', verb: 'DROP' }`.
8. **Unknown treated as destructive.** Call `executeQuery` with `ZZZZ not_a_keyword`. Expect: modal appears (unknown → approval required), classification `{ kind: 'unknown' }`.
9. **Readonly kill-switch.** Mark a connection `readonly`. Try any non-read via MCP. Expect: `{ kind: 'readonly_blocked', classification: { kind: 'write', ... } }` — no modal, no prompt, no execution.
10. **Standalone stdio MCP.** Launch `dist/mcp-server.js` and call `execute_query` with `INSERT INTO t VALUES (1)` against a connection without `agentWriteApproval: 'never'`. Expect: structured error with `kind: 'approval_required'` and a message telling the user to set `'never'` if they really want this CLI to write. Then set `'never'` via the VS Code extension, reload the standalone via `reload_connections`, retry — the write runs.
11. **Classifier edge cases (covered by unit tests).** Leading `-- comment\n` + `DELETE` classified as `write`. `SELECT ';'; UPDATE t SET x=1` classified as `write` (two statements). `INSERT INTO t VALUES (1); SELECT 1` → `write`.

## Checklist
- [x] **CHANGELOG.md** — `[Unreleased] / Added` entry with `(#75)` link
- [x] **CLAUDE.md** — new "MCP Agent Safe Mode" section documenting `queryRisk.ts`, `agentWriteApproval`, and how each MCP surface honors it
- [x] **Unit tests** — `src/test/queryRisk.test.ts` (56 tests): verb classification, comment/whitespace/case handling, quote/backtick-aware split, multi-statement severity, approval-mode gating, describe output
- [ ] **README.md** — not touched; MCP agent safety belongs in the wiki "Agent safety" page once more pieces of #75 land
- [ ] **l10n** — the new modal text (`"Run"`, `"Run & remember for session"`, detail) uses VS Code's own modal strings; no new `package.nls.json` entries
- [ ] **Wiki** — follow-up "Agent safety: write approval" page once EXPLAIN reuse / audit log / shield icon land (not pushed from this PR)


Siyet (commented):
## Fixes pushed
- `8f99014` — Classify `EXPLAIN ANALYZE <DML>` and `WITH ... (DML ... RETURNING)` under the inner destructive verb. Both shapes were being classified as `read` and silently bypassing the readonly kill-switch and the agent approval gate.

## Issues to address
- `src/mcp/server.ts:82-105` — the approval check in `viewstor.mcp.executeQuery` races: two concurrent calls both see `hasSessionApproval === false`, both pop a modal. Low-impact (user sees two modals, each still gated), but worth serialising with an in-flight Promise keyed like `sessionApprovals`.
- `src/mcp/server.ts:89-94` — modal text (`"Run"`, `"Run & remember for session"`, the `detail` body, the main message) is hardcoded English. These are genuine new UI strings, not VS Code built-ins; they should go through `vscode.l10n.t(...)` like every other user-facing message in this repo (e.g. `connectionForm.ts:87`). The PR checklist claims "no new `package.nls.json` entries" but that's only true because the strings aren't wired up for localisation at all.
- `src/mcp-server/index.ts:252-259` — the `add_connection` handler destructures `type` as `'postgresql' | 'redis' | 'clickhouse'`, missing `'sqlite'`. Pre-existing in this file, but it's now visibly stale next to the PR's approval gate — SQLite connections added this way will still classify and gate correctly but the type annotation lies. Tighten to match the schema's `enum`.
- `src/mcp/queryRisk.ts` — transaction-control statements (`BEGIN`, `START TRANSACTION`, `COMMIT`, `ROLLBACK`, `SAVEPOINT`) classify as `unknown` and therefore always prompt. Arguably correct in isolation, but an agent wrapping a write in `BEGIN; ... ; COMMIT` now gets three modals instead of one. Consider treating bare transaction control as `admin`, or as the severity of the most severe inner statement.
- `src/mcp/server.ts:8` — `sessionApprovals` is a module-level `Map` that outlives everything; entries for deleted connections never get purged. Minor memory hygiene — clear on `connectionManager.onDidChange` when a connection is removed.

## Needs maintainer
- Session approval cache keyed by `(connectionId, risk.kind)` rather than SQL text is deliberate per the issue, but it does mean approving one `DELETE` covers all subsequent `DELETE`/`INSERT`/`UPDATE` within 5 min on that connection (since all three share `kind: 'write'`). If the intent was tighter granularity, the key would need to include `risk.verb`. Not fixing — matches the spec in the PR body.
- Default approval is `'always'` even for pre-existing connections with `agentWriteApproval: undefined`. This is a behavioral change agents depending on MCP writes will notice immediately. The PR author called this out in Assumptions; flagging so the maintainer explicitly signs off on the break before merge.

---

PR #97, opened 2026-04-16
MCP: agent anonymization at the boundary (#72)

### LINKED ISSUE #72: Anonymize DB data for LLM agents (per-connection setting)
## Context

Issue #71 controls **whether** an agent can reach a connection. This issue is orthogonal: **if** the agent can execute queries, we want an option that strips PII / sensitive values out of the rows before they leave the MCP boundary.

### Current state audit

I audited what an agent can already see through the UI. Findings (with file:line refs):

- \`src/mcp/server.ts:220\` \`viewstor.mcp.openQuery\` — only opens the editor, returns \`{ success, message }\`, no rows.
- \`src/mcp/server.ts:248\` \`viewstor.mcp.openTableData\` — only opens the panel, returns \`{ success, message }\`, no rows.
- \`src/views/resultPanel.ts\` — \`panels\` Map is private, no public getter. Data lives inside the webview DOM, inaccessible to MCP.
- \`src/commands/shared.ts:56\` \`queryResults\` Map — stores only \`{ line, text, isError, timestamp }\` for CodeLens, no row data.
- Query history (\`historyDocMap\`) — maps id → doc URI, not to results.

**Conclusion:** today an agent CANNOT read data from already-open UI panels or from any cached state. It must call \`execute_query\` / \`get_table_data\` / \`get_table_info\` / \`build_chart\` — all of which return raw values. That is the single chokepoint we need to put anonymization in front of.

This also means \"block UI access\" doesn't need new code, but we should add a test/invariant: no future MCP tool may return rows from an open panel unless it goes through the anonymization filter.

### 1. Per-connection setting

\`\`\`ts
ConnectionConfig.anonymizeForAgents: boolean
// or for future-proofing:
ConnectionConfig.agentAnonymization: 'off' | 'heuristic' | 'strict'
\`\`\`

Also bubble up through folder inheritance (like \`readonly\` / \`color\`) and ties into the \`profile\` preset from #71 — \`prod\` turns it on by default.

### 2. Where to enforce

Single chokepoint: a helper \`anonymizeRows(columns, rows, policy)\` applied at the MCP boundary in both:

- \`src/mcp/server.ts\` — wrap the return of \`executeQuery\`, \`getTableData\`, \`visualize\`, \`getTableInfo\` (sample values), and any future tool that returns rows.
- \`src/mcp-server/index.ts\` — same for \`execute_query\`, \`get_table_data\`, \`build_chart\`, \`export_grafana_dashboard\`.

The filter runs after the driver returns rows, before the MCP response is serialized. That way driver code stays agnostic and future tools get the filter for free as long as they pipe through the helper.

### 3. Detection strategies

**Heuristic (default when on):**
- Column-name patterns (case-insensitive): \`email\`, \`e_mail\`, \`phone\`, \`tel\`, \`mobile\`, \`ssn\`, \`passport\`, \`pass(port|word)\`, \`iban\`, \`card\`, \`cvv\`, \`token\`, \`secret\`, \`api_?key\`, \`auth\`, \`addr(ess)?\`, \`first_?name\`, \`last_?name\`, \`full_?name\`, \`dob\`, \`birth\`.
- Value-shape heuristics for \`text\` columns: email-like regex, phone-like, credit-card Luhn check.
- User-extensible allow/deny lists via settings (\`viewstor.anonymization.sensitiveColumns\`, \`viewstor.anonymization.safeColumns\`).

**Strict:**
- All \`text\` / \`varchar\` / \`json\` / \`jsonb\` / \`bytea\` columns masked regardless of name.
- Numeric, boolean, timestamp, enum values pass through (aggregatable, low PII risk — matches the common \"BI analyst\" threat model).

### 4. Transformations

For each masked cell, configurable strategy per-connection:

- \`null\` — replace with \`NULL\` (simplest, breaks charts that depend on non-null)
- \`hash\` — SHA-256 truncated to 8 chars, deterministic so JOINs on masked keys still work across calls (\`hash('alice@example.com') === hash('alice@example.com')\`)
- \`shape\` — preserve format (\`a@b.com\` → \`x@y.com\`, \`+7 911 123-45-67\` → \`+0 000 000-00-00\`, 16-digit card → 16 x's)
- \`redacted\` — replace with literal \`'<redacted>'\`

Default: \`hash\`. Shape-preserving is a nice-to-have for v1.

### 5. Schema/metadata handling

- \`getTableInfo\` returns column list + types — this is fine to share (no values). Don't anonymize.
- But if we ever add \"sample values\" to \`getTableInfo\` (handy for agents to learn data shape), those MUST go through the filter.
- Error messages: drivers sometimes include offending values in constraint errors (\"duplicate key (alice@example.com) violates...\"). Scrub known sensitive patterns from error \`.message\` too.

### 6. UI

- Connection form: \"Anonymize data returned to agents\" toggle + strategy dropdown (null/hash/shape/redacted) + \"Show sensitive column rules\" link.
- Tree view: add a small mask icon next to the lock icon from #71 for connections with anonymization on.
- Settings page: list of default sensitive column patterns, user can add/remove.

### 7. What NOT to do

- Don't try to enforce anonymization in the **user-facing** panels. The user is the trusted operator; they see raw data. This is strictly an agent-boundary concern.
- Don't rewrite the user's SQL (brittle, vendor-specific). Filter the result rows.
- Don't claim this is a full DLP solution. Document it as \"best-effort masking to reduce accidental PII exposure to LLMs\". Determined exfiltration via aggregation / side channels is out of scope.

### 8. Interaction with other features

- \`build_chart\` / \`export_grafana_dashboard\` — the chart preview data and Grafana dashboard JSON go through the same filter. Aggregated / bucketed queries usually don't carry PII, so in practice most charts look normal.
- Copilot Chat (\`@viewstor\`) — schema context (table/column names + types) is fine. If a future \`/data\` slash command is added, it MUST pipe through the filter.

## Acceptance criteria

- [ ] \`anonymizeForAgents\` (or \`agentAnonymization\`) field on \`ConnectionConfig\` with folder inheritance
- [ ] Form toggle + strategy dropdown
- [ ] \`profile: 'prod'\` (from #71) turns anonymization on by default
- [ ] Single helper \`anonymizeRows(columns, rows, policy)\` in \`src/mcp/anonymizer.ts\` (pure, vscode-independent, unit-testable)
- [ ] Applied in both in-process MCP (\`src/mcp/server.ts\`) and standalone MCP (\`src/mcp-server/index.ts\`)
- [ ] Heuristic column-name detection with extensible settings-based allow/deny list
- [ ] Four transformations: \`null\`, \`hash\`, \`shape\`, \`redacted\`
- [ ] Error-message scrubbing for driver errors returned to agents
- [ ] Unit tests: per-strategy × per-column-type matrix; JOIN determinism for \`hash\` mode; regex patterns for shape-preserving emails/phones/cards
- [ ] Integration test: turn on anonymization, run \`execute_query SELECT email, name, age FROM users\` via MCP → verify email is hashed, name depends on policy, age passes through
- [ ] Regression guard: assertion in code (+ test) that no MCP tool returns row data bypassing \`anonymizeRows\`
- [ ] Docs: README section + wiki page explaining the threat model, strategies, and limitations

## Out of scope

- Per-column explicit rules in the UI (Settings-based regex is enough for v1; follow-up issue for per-column toggles in the schema tree)
- Differential-privacy style aggregation limits — separate concern
- Sanitizing user-facing panels — only affects the agent boundary

Siyet:
## Summary

Optional PII masking for rows returned by MCP tools (Claude Code, Cursor, VS Code agents). Applied at **both** MCP surfaces — VS Code in-process commands (`viewstor.mcp.executeQuery` / `getTableData` / `visualize`) and the standalone stdio server (`execute_query` / `get_table_data` / `build_chart`) — after drivers return rows and before responses are serialized, so drivers stay agnostic and new tools pick the filter up for free.

### Modes
- `off` — no masking (default)
- `heuristic` — mask columns whose names match `email` / `phone` / `ssn` / `iban` / `token` / etc. Names are normalized (`user_email` → `user email`) so underscores aren't swallowed by JS `\b`.
- `strict` — mask every text-like column (`text`, `varchar(*)`, `character varying(*)`, `citext`, `json`, `jsonb`, `bytea`, `blob`, `nvarchar`, `String`, `longtext` …) regardless of column name.

### Strategies
- `hash` — deterministic SHA-256 → 8 hex chars. **JOIN-safe** — the agent can still correlate masked keys across queries.
- `shape` — format-preserving: `alice@example.com` → `x@y.xxx`, Luhn-valid card digits → `x`, phone digits → `0`, generic alphanumerics → `x` with separators preserved.
- `null` — replace cell with `null`.
- `redacted` — replace cell with empty string.

### Configuration
New fields `agentAnonymization` / `agentAnonymizationStrategy` on both `ConnectionConfig` and `ConnectionFolder`. Resolution follows the same folder-inheritance pattern as `readonly` / `color`, with a cycle guard mirroring `moveFolderToFolder`. Connection-level values override folder-level. UI controls added to the Advanced collapsible of the Connection form and to the Folder form.

### Error scrubbing
`scrubErrorMessage` also replaces emails with `[redacted-email]` and Luhn-valid digit runs with `[redacted-card]` in driver error messages, so constraint violations like `duplicate key (alice@example.com) violates…` don't leak raw values.

### Performance
Zero-allocation fast path: when `mode === 'off'` or no columns match, the original rows array is returned by reference (no copy, no mutation).

### Scope deferred to follow-ups
- `profile: 'prod'` preset (depends on #71 credentials encryption)
- Tree-view mask icon to indicate an anonymized connection
- User-extensible settings-based sensitive-column patterns

## Test plan
- [x] 44 new vitest cases cover the per-strategy × per-column-type matrix, JOIN determinism for hash, shape regexes for emails/phones/cards, folder inheritance cycle guard, and error scrubbing
- [x] `npx tsc --noEmit` clean
- [x] `npm run lint` — zero errors (18 pre-existing warnings unchanged)
- [x] `npm test` — 821 tests pass including the new 44
- [x] `npm run build` — clean (only the expected `pg-native` warning)
- [ ] Manual smoke: open Connection form → toggle `strict` + `shape` → run an MCP query → confirm text columns are masked client-side

Closes #72


Siyet (commented):
## Fixes pushed

- `9294e46` — `CARD_DIGIT_REGEX` gained `/g`; a non-Luhn digit run earlier in an error string no longer shadows a real card later in the same message. Outer `catch` in the standalone MCP server now scrubs thrown driver errors using the connection's resolved policy, so exceptions from `execute_query` / `get_table_data` / `build_chart` don't leak PII past the boundary. Regression tests added.

## Issues to address

- `src/mcp/server.ts:132` — `viewstor.mcp.visualize` returns ``` `Connection failed: ${err.message}` ``` without scrubbing. Driver connect errors can include the configured username (and occasionally the host/user blob that agents shouldn't see). Consider routing through `scrubErrorMessage` for parity with the other handlers.
- `src/mcp/server.ts:43,107` — `getSchema` / `getTableInfo` error paths don't scrub. Low risk (catalog queries rarely quote row data), but inconsistent with `executeQuery` / `getTableData` / `visualize`.
- `src/mcp/anonymizer.ts:58-75` — `STRICT_MODE_SENSITIVE_TYPES` doesn't cover ClickHouse `LowCardinality(String)` / `FixedString(N)` or `Enum*(...)`; `isStrictTextColumn` strips the parens, so base becomes `lowcardinality` / `fixedstring` / `enum8` and the column slips past strict mode. Strict mode also treats `numeric` / `decimal` as non-sensitive, which matches intent but is worth pinning in a test so the contract is explicit.
- `src/mcp/server.ts:49-54` — MCP `executeQuery` uses `state?.config.readonly` directly instead of `connectionManager.isConnectionReadonly(id)`, so a read-only folder doesn't gate writes for agents unless the per-connection flag is also set. Pre-existing, but worth noting since the anonymization path correctly uses the folder-aware resolver.
- `src/mcp/anonymizer.ts:157-160` — `isSensitiveColumnName` compares case-insensitive word matches, but the pattern list doesn't include `name` standalone nor `city` / `zip` / `postal` / `street`. Heuristic mode leaves addresses in plain text even when column names are obvious. The PR body already notes settings-based extensible patterns as a follow-up; adding a short TODO comment next to the list would make that connection discoverable from the code.

---

Issue #94, opened 2026-04-15
Webview components: dedupe data table, SQL highlighter, color picker, context menu


Siyet:
## Motivation

After landing the `@vscode-elements/elements` migration ([#86](https://github.com/Siyet/viewstor/issues/86), [#87](https://github.com/Siyet/viewstor/issues/87)) the webview layer still carries several near-duplicates of the same component — most notably **the data table**, which has two parallel implementations. Each new feature (a copy format, a fix for cell-height jumps, a zebra-tint tweak) has to land twice and inevitably drifts. The goal of this issue is to extract these into a single shared layer built on top of `tokens.css` so there is **one table**, one color picker, one context menu, one SQL highlighter.

### 1. Data grid (highest priority — this is what the user asked about)
Two parallel implementations:

| Concern | Result Panel | Diff Panel |
|---|---|---|
| Source | [`src/views/resultPanel.ts`](../blob/main/src/views/resultPanel.ts) (1758 lines) | [`src/diff/diffPanel.ts`](../blob/main/src/diff/diffPanel.ts) + [`src/webview/scripts/diff-panel.js`](../blob/main/src/webview/scripts/diff-panel.js) |
| Drag-select with anchor cell | ✅ | ✅ (ported in #87) |
| `sel-top/bottom/left/right` edge classes | ✅ borders | ✅ box-shadow inset |
| Context menu (Copy / TSV / CSV / MD / JSON) | ✅ | ✅ |
| Ctrl+C → TSV | ✅ | ✅ |
| Zebra striping | ❌ | ✅ |
| Status row tints (added/removed/changed) | ❌ | ✅ |
| Sticky header | ✅ | ✅ |
| Column sort | ✅ | ❌ |
| Inline cell edit | ✅ | ❌ |
| Row-number gutter | ✅ | ❌ |
| Resize selection handle | ✅ | ❌ |

Today improvements to one don't propagate to the other — e.g. the cell-height-jump fix (box-shadow inset) and zebra-with-status in the diff panel don't exist in the Result Grid; the Result Grid's resize handle and column sort don't exist in the diff. That's exactly the drift the user flagged.

### 2. SQL token highlighter
Identical regex \`SQL_KW\` + \`highlightSql()\` body lives in both:
- [`src/views/resultPanel.ts`](../blob/main/src/views/resultPanel.ts)
[...]

PR #98, opened 2026-04-16
refactor(webview): shared color picker for Connection + Folder forms

Siyet:
## Summary

- Extract the swatch + hex textfield + Random/Clear buttons + theme-color palette shared by the Connection form and Folder form webviews into a single `src/webview/scripts/color-picker.js` that installs `window.ViewstorColorPicker` with `hslToHex(h,s,l)`, a frozen `COLOR_PALETTE` (the 12 VS Code terminal ANSI theme colors), and `attach({ textEl, pickerEl, swatchEl, clearBtn?, randomBtn?, paletteEl? }) → { setValue, getValue }`.
- `connection-form.js` and `folder-form.js` now call `ViewstorColorPicker.attach({...})` and use the returned `setValue` / `getValue` handle — replacing two near-identical copies of `hslToHex` + the palette array that had to be hand-synced (the exact drift the parent issue flags).
- `connectionForm.ts` / `folderForm.ts` inject the new `color-picker.js` with its own `<script>` tag before the form-specific script; webpack's `CopyPlugin` already copies `src/webview/scripts/*.js` into `dist/scripts/` so no build-config change is needed.

Part of #94 (step 1: extract cross-panel webview utilities — the issue explicitly calls this out as low-risk and recommends each step be its own PR).

## Manual test cases

1. **Golden path — new connection.** New Connection form → swatch, palette, Random, Clear all still work. Pick a palette swatch → text field shows `var(--vscode-terminal-ansiRed)` and the preview swatch turns red. Click Random → text field + native picker + swatch all update to the same hex. Click Clear → text field empties, picker resets to `#1e1e1e`, swatch returns to transparent.
2. **Hex input → picker sync.** Type `#abcdef` in the text field → native picker and swatch both update to that color. Type `#12` → neither updates until the input becomes a valid hex (or starts with `var(`).
3. **CSS var input.** Type `var(--vscode-terminal-ansiGreen)` in the text field → swatch turns green; native picker stays on its previous hex (it can't hold non-hex values).
4. **Edit existing connection.** Open a saved connection that already has a color. Expected: swatch is primed from the stored value on first render; the palette is populated once; editing the value propagates as in (1)–(3).
5. **Folder form.** Repeat (1)–(4) on the Folder form — same behavior, same palette, same Random/Clear.
6. **Save round-trip.** Pick a color, save, reopen: `colorPicker.getValue()` returns the trimmed value, persisted correctly via the existing `config.color` / `folder.color` fields.
7. **Theme compatibility.** Switch VS Code between light, dark and high-contrast while either form is open. Expected: palette buttons pick up `--vscode-terminal-ansi*` values; the "default picker color" (`#1e1e1e`) is only ever written into the native `<input type="color">` when the text field is empty, so the text field never shows it.

## Checklist

- [x] `npm run lint` — 0 errors (18 pre-existing warnings unchanged)
- [x] `npm test` — 804 pass in CI mode; new 27 tests in `colorPicker.test.ts` (pure `hslToHex`, palette shape, full attach/setValue/getValue DOM wiring via `vm` sandbox)
- [x] `npm run build` — clean (only the pre-existing `pg-native` warning); `dist/scripts/color-picker.js` is copied in alongside the existing webview scripts
- [x] CHANGELOG.md updated under `[Unreleased] > Changed` with `(#94)` link
- [x] CLAUDE.md updated under **Webview UI foundations** to document the new `color-picker.js` module
- [x] README.md — not touched; the color picker is internal webview UI, unchanged behaviorally
- [x] l10n — no new UI strings
- [x] Wiki — no new page needed; this is a refactor


Siyet (commented):
## Fixes pushed

- `3dc0c77` — drop stale docstring reference to non-existent `src/webview/colorPicker.ts` in `src/webview/scripts/color-picker.js`; point at the real test file instead.


---

PR #100, opened 2026-04-16
feat(mcp): per-connection AI agent access mode (#71)

### LINKED ISSUE #71: Restrict LLM/MCP access per connection (hide from agents, encrypted credentials)

Siyet:
Closes #71 (first slice).

Introduces a per-connection / per-folder `agentAccess` field that gates what VS Code chat and MCP tools can do with a connection.

## Modes

| mode          | `list_connections` | `get_schema` / `get_table_info` | `execute_query` / `get_table_data` / `build_chart` / `export_grafana_dashboard` / `visualize` | `openQuery` / `openTableData` |
| ------------- | :----------------: | :-----------------------------: | :-------------------------------------------------------------------------------------------: | :---------------------------: |
| `full`        |         ✓          |                ✓                |                                               ✓                                               |               ✓               |
| `schema-only` |         ✓          |                ✓                |                                               ✗                                               |               ✗               |
| `none`        |    hidden          |         "not found"             |                                          "not found"                                          |          "not found"          |

`none` is indistinguishable from a missing connection so agents cannot even discover it.

## Resolution order

1. Connection's own `agentAccess`
2. Walk up the folder chain via `parentFolderId` — nearest ancestor with `agentAccess` wins (cycle-guarded)
3. `viewstor.defaultAgentAccess` setting (default `full`, keeps existing MCP users unaffected)

Implemented once as a pure helper in `src/mcp/agentAccess.ts` and reused by both `ConnectionManager` (VS Code) and `ConnectionStore` (standalone stdio MCP).

## UI

- Connection and Folder forms gain an `AI agent access` dropdown in the Advanced section
- Connections with restricted access show `[agents: schema-only]` / `[agents: none]` after the description in the tree view (plain text — `$(codicon)` doesn't render in `TreeItem.description`)
- New setting `viewstor.defaultAgentAccess` (full | schema-only | none)

## Out of scope (deliberate — keeps the diff reviewable)

These are tracked as follow-ups under #71:

- Credential encryption at rest in `~/.viewstor/connections.json`
- Per-profile presets (e.g. "Development", "Production read-only")
- Credential isolation for the standalone MCP server (separate config path / sandbox)

## Test plan

- [x] `npm run lint` — 0 errors (warnings are pre-existing)
- [x] `npm test` — all 799 tests green; 22 new tests cover the pure helper, `ConnectionManager.getAgentAccess`, and standalone `ConnectionStore.getAgentAccess` including folder-chain walking and cycle guard
- [x] `npm run build` — webpack succeeds
- [ ] Manual: open connection form, set `none` on a connection, confirm MCP `list_connections` omits it and `execute_query` returns "not found"
- [ ] Manual: set a folder to `schema-only`, verify all nested connections inherit; override on a child connection wins
- [ ] Manual: verify tree badge renders; verify legacy connections (no `agentAccess` field) still behave as `full`

https://claude.ai/code/session_01DNrRcxZcYDeEcBWfCJJozs


Siyet:
Heads-up: the E2E Tests job on this PR came back as failed after 51s (exit code 1, no detail surfaced in the annotation). Looks like a flake or transient runner issue rather than a real regression:

- The diff only touches TypeScript types (additive `agentAccess?` on `ConnectionConfig` / `ConnectionFolder`), a new pure helper (`src/mcp/agentAccess.ts`), `ConnectionManager` / standalone `ConnectionStore` methods, the MCP gate in both surfaces, form UI, and a tree badge. No driver or testcontainer code is touched.
- Of the files `src/test/e2e/**` imports from my diff, the only one is `src/types/connection.ts` — and the only change there is one optional field. No runtime behaviour change.
- Same parent commit on trunk (#92) passed E2E in 50s; this run failed in 51s — similar timing points away from a Docker-startup issue.
- Unit Tests (844 pass locally), Lint, Build, Security Audit, and Changeset Size all green.

Could you re-run the E2E job when you get a chance? Happy to push an empty commit to retrigger if that's easier.

---

Siyet:
## Review

Solid first slice — the pure helper + dual-surface gate is the right shape. The folder-chain walk with cycle guard is exactly what the issue called for, and reusing one resolver between `ConnectionManager` (VS Code) and `ConnectionStore` (stdio) avoids two implementations drifting.

A few notes:

### Findings

**1. Chat participant was not gated (auto-fixed).**
The PR description says the mode "gates what VS Code chat and MCP tools can do," but `src/chat/participant.ts` was untouched. `none`-mode connections were still surfaced to `@viewstor` (and would be auto-picked as the fallback connection), and `/chart` would happily run SQL on `schema-only` connections — defeating exactly the gate the MCP path enforces. Fixed in dfaa2f3 (see "Fixes pushed" below).

**2. `ConnectionStore.getAgentAccess` returning `'none'` for an unknown id is a deliberate divergence from `ConnectionManager` (which returns the default).** That's a useful safety property — and the test `"returns 'none' for unknown connection so MCP treats it as missing"` documents the intent — but consider a one-line block comment on `getAgentAccess` itself flagging the divergence. Future readers will compare the two implementations and wonder. Not blocking.

**3. Pre-existing, not introduced by this PR but worth a follow-up:** `folderForm.setFolder` message handler doesn't restore `scope`, and `updateFolder`'s `Pick<>` type omits `scope`. Easy to miss when adding new fields. Mentioning here so it gets a separate issue rather than being dragged into this one.

### Out-of-scope items I didn't touch
- Credential encryption / isolation — already explicitly deferred on the PR
- The "set a folder to `schema-only`, verify all nested connections inherit" manual checklist item — covered by the `inherits from immediate folder` and `walks up multiple folder levels` unit tests; safe to tick

### Test plan
Local gate after the chat fix:
- `npx tsc --noEmit` — clean
- `npm run lint` — 0 errors (18 pre-existing warnings)
- `npm test` — 799 passed (the 1 unhandled tinypool worker error is a pre-existing flake from `connectionStore.test.ts` reloading `better-sqlite3` via `vi.resetModules()`; unrelated to this PR)
- `npm run build` — webpack succeeds (only pre-existing `pg-native` warning)
- All 6 CI checks green on the new commit

## Fixes pushed

- dfaa2f3 `fix(chat): gate Copilot Chat participant by agentAccess mode`
  - `resolveConnectionId()` now skips `none` connections at every fallback step (active-editor URI, first-connected, first-existing) so a hidden connection is never auto-picked
  - Defensive guard at handler entry: if access resolves to `none`, surface the same "no active connection" message as if it didn't exist
  - `/chart` is refused for `schema-only` connections before we burn an LLM round-trip; `/schema` and `/describe` keep working since they only read metadata


---

Issue #74, opened 2026-04-14
Diff: panel-first UX, add multiple tables, validate cross-DB-type comparison


Siyet:
## Context

Two related rough edges in the data-diff feature today:

1. **Compare With...** opens a QuickPick before showing anything. The user has to pick the second table blind — they can't see what they picked the first one for, can't change their mind without restarting. Feels backwards.
2. The diff is hardcoded to **exactly two sources**. There's no way to add a third for \"compare prod / staging / dev side-by-side\" or to drop one and pick another without closing the panel and starting over.

Also, while #67 added the Statistics tab, it's only enabled when both sides are the same database type (\`leftState.config.type === rightState.config.type\`). Cross-type comparisons (PG ↔ ClickHouse, SQLite ↔ PG for migration validation) work for rows and schema but lose the Statistics tab entirely. Worth fixing — cross-type stats can still be useful if we agree on a common subset.

### 1. Panel-first UX

Right-click table → **Compare With...** → diff panel opens immediately with the source table loaded as the **left** pane and the right pane in an empty \"placeholder\" state:

\`\`\`
┌─────────────────────────┬─────────────────────────┐
│ users — prod (loaded)   │  + Pick a table         │
│ 12,847 rows             │     to compare with     │
│                         │                         │
└─────────────────────────┴─────────────────────────┘
\`\`\`

Clicking the placeholder opens the existing table-picker QuickPick. The picker is now reachable from inside the panel — so the user keeps context while choosing.

Same for the command-palette entry **Viewstor: Compare Data**: open empty panel, both panes are placeholders, user fills them in.

### 2. Multi-pane (N-way) diff

Add an \"+\" button at the top right of the panel to add another source:

\`\`\`
┌──────────────┬──────────────┬──────────────┬───┐
│ users — prod │ users — stg  │ users — dev  │ + │
└──────────────┴──────────────┴──────────────┴───┘
\`\`\`

Each pane has an \"x\" to remove it. Minimum 1 pane (otherwise close the panel), maximum reasonable (e.g. 5 to keep the UI sane).

How each tab handles N panes:

- **Row diff**: pick a designated \"left\" pane as the key source, every other pane is matched against it. Display: one column per pane for non-key fields, key columns shown once. Cells highlight when they differ from the left.
- **Schema diff**: existing column comparison extends to N columns; status column shows per-pair diffs (\"left/staging differ in type, left/dev same\").
[...]

PR #103, opened 2026-04-16
feat(diff): panel-first UX for Compare With...

Siyet:
## Summary

- Right-click `Compare With...` on a table now opens the diff panel **immediately** with the source table loaded on the left and a dashed "Pick a table to compare with" placeholder on the right.
- The existing `pickTableWithLoading` quickpick auto-opens on top of the panel; dismissing the picker leaves the placeholder clickable so the user can re-invoke it from inside the panel (acceptance bullet 1 from #74).
- Once a right table is chosen, `DiffPanelManager` **adopts the same webview panel** via a new optional `existingPanel` parameter on `show()`, so the user sees the full Row/Schema/Stats diff render in place — no second tab opens, no flicker.

Part of #74 (panel-first UX slice). Closes the first acceptance bullet; N-way panes and the `Viewstor: Compare Data` command-palette equivalent remain for follow-up PRs.

## Assumptions

- This PR ships **only** the `Compare With...` path of the panel-first UX. The command-palette `Viewstor: Compare Data` flow still uses the historical two-picker chain and is left untouched to keep the diff focused — a follow-up can generalize the placeholder pane to either side.
- The placeholder uses `<vscode-icon name="arrow-both">` and a native `<button>` styled with existing design tokens rather than a `<vscode-button>` because the pending HTML runs before the elements bundle has finished booting; a plain button avoids the pre-upgrade click-miss window and matches the dashed "empty state" styling described in the issue mock.
- Tests: the existing `Diff Panel (vscode-elements)` suite exercises `mgr.show(left, right, options)` directly with no `existingPanel` argument — the new trailing optional parameter is backwards-compatible so that suite stays green. No new unit test added for the pending panel because it's pure-UI plumbing; the end-to-end behaviour (placeholder → picker → diff) is exercised by the manual test plan below.

## Manual test cases

1. **Golden path — Compare With...**
   - Connect two PG databases, right-click `public.users` on connection A → `Compare With...`.
   - **Expected:** diff panel opens immediately showing `A → users` on the left and a dashed "Pick a table to compare with" card on the right. The quickpick opens on top.
   - Select `public.users` on connection B → picker closes → the same panel re-renders with the full Row/Schema tabs. No second tab is created.

2. **Dismiss picker, re-pick from inside**
   - Right-click `public.users` → `Compare With...` → picker opens with the panel visible behind it.
   - Press `Esc` to dismiss the picker without choosing.
   - **Expected:** panel stays open with the placeholder still visible. Click the "Pick a table to compare with" button → picker re-opens. Select a table → diff renders in place.

3. **Cross-connection compare**
   - Right-click a table on a PG connection → `Compare With...` → select a table on a SQLite connection.
   - **Expected:** panel opens immediately with PG label on left + placeholder; after picking the SQLite table the full diff renders with Row + Schema tabs (Statistics tab absent because drivers differ, per existing behaviour).

4. **No-PK edge case**
   - Right-click a table that has no primary key → `Compare With...` → pick a second table.
   - **Expected:** panel-first placeholder shows → after picking, the existing "select key column(s)" quickpick appears → diff renders.

## Checklist

- [x] README — "Data Diff" bullet updated to describe the panel-first flow.
- [x] CHANGELOG — new Unreleased entry under `### Changed`.
- [ ] CLAUDE.md — no new modules; existing Data Diff section already describes `DiffPanelManager`, the new `showPending` method is a private-ish surface not worth calling out.
- [ ] l10n — picker placeholder string reused unchanged (`vscode.l10n.t('Select table to compare with "{0}"', …)`).
- [ ] Wiki — `Data-Diff.md` should be updated once the remaining #74 slices (N-way, command-palette panel-first) land; noting here so follow-up PRs pick it up.


Siyet (commented):
## Fixes pushed

- `5ab5ac6` — Guard against disposed pending panel in `compareWith`. If the user closes the pending panel while the table picker is open or data is loading, `runPicker` now bails early instead of throwing when `show()` tries to set properties on the disposed webview.

---

PR #110, opened 2026-04-24
feat: harden cross-database table comparison

Siyet:
## Summary

- Normalize the presence of `row_count`, `total_size`, and `last_modified` across PostgreSQL, ClickHouse, SQLite, and Redis while preserving unavailable values as `null`.
- Keep cross-type Statistics semantically safe: only explicitly comparable metrics are shown (`row_count` by default); incompatible storage and engine-specific metrics remain hidden and are explained in the UI.
- Make cross-database row comparison type-aware for exact numerics, PostgreSQL booleans versus SQLite/ClickHouse `0`/`1`, and equivalent UTC timestamp representations.
- Add a column-mapping picker for unequal schemas: exact pairs are selected by default, similar-name pairs can be selected explicitly, and one-sided fields stay out of Row Diff unless requested.
- Resolve compatible primary keys from either side, including ClickHouse MergeTree keys from `system.columns`, and deduplicate repeated PK metadata.
- Add synchronized search to Row Diff and preserve it across filters and query reruns.
- Include SQLite tables and views in Compare With, surface data-loading errors instead of opening empty diffs, and preserve optional metadata degradation.
- Return exact view row counts when PostgreSQL estimates or ClickHouse `system.tables.total_rows` are unavailable; failed count queries remain unavailable rather than becoming a misleading zero.
- Make the macOS development build detect current VS Code Electron versions and compile `better-sqlite3` from source for the host architecture when no prebuild exists.

Part of #74

## Statistics semantics

`COMMON_STAT_KEYS` is a driver presence contract, not proof that values are comparable across engines. PostgreSQL relation size, ClickHouse compressed storage, SQLite page usage, and Redis memory allocation are different physical concepts even though all use bytes.

## Validation

- 133 targeted unit/integration tests passed for diff engine, column selection, commands, panel lifecycle, and ClickHouse statistics.
- 26 real-driver Docker E2E scenarios passed across PostgreSQL, ClickHouse, and SQLite, covering row/schema/statistics comparison, table-to-view mappings, booleans, timestamps, view row counts, hidden cross-type metrics, and PK behavior.
- ESLint, TypeScript test compilation, production webpack build, and diff checks passed.
- Manually verified local PostgreSQL, ClickHouse, and SQLite fixtures in the Extension Development Host.

## Documentation

- CHANGELOG, CLAUDE.md, AGENTS.md, and the Compare Tables E2E matrix are updated.


Siyet (commented):
## Fixes pushed

- `20ab191` — Handle Redis `stream` type in `getTableStatistics` (was returning `null` row count for streams; now uses `XLEN`). Collapse SQLite index-size N+1 loop into a single `dbstat` subquery.

## Issues to address

- `src/drivers/redis.ts:5` — `TableStatistic` is imported but `COMMON_STAT_KEYS` is not. The constant is exported for downstream tooling per the PR description, but the driver itself doesn't reference it, so no functional impact — just noting the asymmetry with the test file which does import it.

---

PR #111, opened 2026-04-24
feat(diff): panel-first UX for Compare Data command palette

Siyet:
## Summary
- `Viewstor: Compare Data` from the command palette now opens the diff panel immediately with two dashed placeholder "Pick a table" panes instead of requiring both tables to be chosen upfront via sequential QuickPicks
- Clicking a pane opens the table picker (QuickPick); once both sides are filled, the panel transitions in-place to the full Row/Schema/Stats diff view by reusing the same webview panel
- `DiffPanelManager.showPending(pickTable)` is the new entry point; `show()` refactored to `showInPanel()` internally to accept an optional existing panel for the transition

Part of #74 (command-palette panel-first UX slice). The context-menu `Compare With...` flow is unchanged (PR #103 handles that independently).

## Assumptions
- The pending panel uses native `<button>` elements rather than `<vscode-button>` to avoid the pre-upgrade click-miss window before the web components bundle finishes booting.
- Only the command-palette `compareData` flow is changed. The `compareWith` context-menu flow stays with its existing sequential UX — PR #103 handles that path independently.
- No new l10n strings added; reuses existing `Select LEFT table` / `Select RIGHT table` via `vscode.l10n.t()`.

## Manual test cases
1. **Golden path** — Command Palette → `Viewstor: Compare Data` → panel opens immediately with two dashed placeholder panes and an arrow-both icon between them. Click the left "Pick a table" → QuickPick opens showing all tables from connected databases → select one → left pane updates to show the connection + table name with a check icon. Click the right placeholder → pick a table → "Loading comparison data…" appears → full diff (Row/Schema/Stats tabs) renders in place. No second panel tab created.
2. **Dismiss picker, re-pick** — Open Compare Data → click left placeholder → QuickPick opens → press Escape without choosing → panel stays with both placeholders. Click left again → pick a table → left fills. Click right → pick → diff loads.
3. **No connected databases** — Disconnect all databases → Command Palette → Compare Data → warning "No connected databases. Connect first." Panel does not open.
4. **No-PK table** — Pick two tables where the left has no primary key → after both are selected, the "select key column(s)" QuickPick appears → select columns → diff renders.
5. **Cross-connection** — Pick a PG table on the left, a SQLite table on the right → diff renders with Row + Schema tabs (Statistics tab may be absent if drivers differ, per existing behavior).

## Checklist
- [x] README — updated Data Diff section with panel-first description
- [x] CHANGELOG — new Unreleased entry under `### Changed`
- [x] CLAUDE.md — updated Data Diff section with `showPending` method description
- [ ] l10n — no new strings
- [ ] Wiki — to be updated alongside the remaining #74 slices

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jw1DXScrzu9XPgspdiBRzx)_


Siyet (commented):
## Fixes pushed

- `034ddc7` — Handle stuck loading state and panel disposal race in `transitionPendingToFull`. Added `resetPendingPanel()` helper that clears picks and rebuilds the pending HTML when an early return occurs (driver unavailable, user cancels key-column QuickPick, or any thrown error). Added staleness checks (`this.pendingDiff !== state`) after each async boundary to bail out cleanly when the user closes the panel while data is loading.

---

PR #112, opened 2026-04-24
feat(diff): N-way multi-source diff engine

### LINKED ISSUE #74: Diff: panel-first UX, add multiple tables, validate cross-DB-type comparison

Siyet:
## Summary
- Add `computeNWayRowDiff`, `computeNWaySchemaDiff`, and `computeNWayStatsDiff` — pure functions that generalize the existing 2-source diff to N sources (up to 5 panes per issue spec)
- Row diff uses a star topology: one reference source (configurable via `referenceIndex`) compared against all others by key columns, with `values[]` tuples per matched row and `sourceOnly[][]` for per-source unique rows
- `nwayToLegacyRowDiff()` adapter converts N-way results back to the 2-source `RowDiffResult` shape so existing call sites and webview code continue working unchanged

Closes #74

## Assumptions
- **Star topology for N-way row diff.** The issue specifies "pick a designated 'left' pane as the key source, every other pane is matched against it" — implemented as `referenceIndex` (default 0). Pairwise N×N comparison was considered but rejected for complexity and UX clarity.
- **Engine layer only in this PR.** The webview rendering of N columns, "+" / "x" pane buttons, and the `pickTable` webview message flow are left for follow-up PRs that build on these engine types. The existing 2-source panel and all call sites are unaffected.
- **N-way stats diff uses presence flags, not delta/percent.** The 2-way `StatsDiffItem` has `delta` and `deltaPercent` which only make sense for exactly 2 sides. The N-way `NWayStatsDiffItem` uses `values[]`, `present[]`, and `hasDifferences` instead — the webview can compute pairwise deltas on demand.

## Manual test cases
- Run `npm test` — all 944+ existing tests pass unchanged, plus new N-way tests
- Import `computeNWayRowDiff` in a scratch test and verify 3-source diff with mixed add/remove/change/unchanged rows
- Import `computeNWaySchemaDiff` with 3 column sets where one source has an extra column — verify `present[]` flags and `hasDifferences`
- Call `nwayToLegacyRowDiff(computeNWayRowDiff([left, right], opts))` and verify the result matches `computeRowDiff(left, right, opts)` for the same inputs

## Checklist
- [x] README — added N-way diff engine bullet in Data Diff section
- [x] CHANGELOG — added entry under [Unreleased] → Added
- [x] CLAUDE.md — updated Data Diff section with N-way function descriptions and types
- [ ] Wiki — N-way diff documentation (note in PR body, not pushed)

---
_Generated by [Claude Code](https://claude.ai/code/session_01RAZNLgt7xKJc6SuMKe5Mqu)_


Siyet (commented):
## Fixes pushed

- `114a09b` — Fix inaccurate `sourceOnly` docstring in `NWayRowDiffResult`: for non-reference sources, rows sharing a key across multiple non-ref sources appear in all their `sourceOnly` arrays, contradicting "only in source i and no other". Updated to "rows from source i not present in the reference source".
- `114a09b` — Rename shadowed variable `n` → `v` in `computeNWaySchemaDiff` arrow function (`nullables.some(n => ...)` shadowed outer `const n = columnSets.length`).
- `114a09b` — Add runtime guard in `nwayToLegacyRowDiff`: throws if `sourceCount !== 2`, since `values[1]` can be null for matched rows when N>2 (ref present + one other present doesn't guarantee source 1 is present). Added corresponding test case.

## Issues to address

- `src/diff/diffEngine.ts:636` — `added` increment uses a ternary that is always truthy in the `!refRow` branch (the key must exist in at least one non-reference source to be in `allKeys`). Simplify to `added++` for clarity.
- CLAUDE.md N-way diff section should mention `nwayToLegacyRowDiff` throws for N≠2, since downstream consumers need to know the constraint.

---

PR #118, opened 2026-04-28
feat: add MySQL / MariaDB driver

### LINKED ISSUE #8: feat: add MySQL / MariaDB driver
## Problem

MySQL and MariaDB are among the most widely used relational databases. Users migrating from DBeaver or DataGrip expect MySQL support out of the box.

## Proposed solution

Add a `MysqlDriver` implementing `DatabaseDriver` interface.

- **Library:** [`mysql2`](https://www.npmjs.com/package/mysql2) (Promise API, prepared statements, TLS)
- **Protocol:** TCP (default port 3306)
  - `getSchema()` → `SHOW DATABASES` + `information_schema.tables` + `information_schema.columns`
  - `getDDL()` → `SHOW CREATE TABLE`
  - `cancelQuery()` → `KILL QUERY <id>` on separate connection
  - `getEstimatedRowCount()` → `information_schema.tables.TABLE_ROWS`
  - `getCompletions()` → tables, columns, schemas from `information_schema`
  - Safe mode / readonly fully supported
- **Connection form:** add `mysql` to `DatabaseType`, default port 3306, charset selector (utf8mb4)
- **Tests:** e2e with `testcontainers` (`mysql:8` image)

## Alternatives considered

- `mysql` (callback-based, older) — `mysql2` is the maintained successor
- `knex` / `prisma` — too heavy, we want raw driver access

Siyet:
## Summary
- Add `MysqlDriver` implementing the full `DatabaseDriver` interface via `mysql2` (Promise API, connection pooling, TLS)
- All required + all optional methods: schema browsing (`information_schema`), DDL (`SHOW CREATE TABLE/VIEW/TRIGGER`), safe mode (`EXPLAIN` → `ALL` detection), cancel (`KILL QUERY`), estimated row counts, enum/set autocomplete, table statistics, indexes/constraints/triggers for schema diff, indexed column tracking
- Connection form, import service (DBeaver/DataGrip), MCP server, chart time bucketing (`DATE_FORMAT()`), and safe mode all updated for MySQL

Closes #8

## Assumptions
- MySQL uses backtick identifier quoting. The driver uses its own `quoteId()` helper instead of the shared `quoteIdentifier()` which produces double-quoted identifiers. The `buildAggregationQuery` in `chart.ts` was updated with MySQL-aware quoting. `buildFullDataQuery` and `queryHelpers.ts` SQL builders still use double quotes — inline editing features will need MySQL-specific SQL generation in a follow-up.
- The icon font (`viewstor-icons.woff2`) does not yet contain a MySQL glyph at `\E005`. The tree will show a placeholder until the font is regenerated with the MySQL icon.
- `cancelQuery()` uses `KILL QUERY` on the current connection ID obtained via `SELECT CONNECTION_ID()`. In a pooled environment this cancels the query on the connection that runs the cancel, not necessarily the one running the target query. A more robust approach would track the connection ID at query start — deferred to follow-up.

## Manual test cases
- **Golden path**: Add MySQL connection → connect → browse schema → run SELECT → view table data → export → disconnect
- **Safe mode**: Run `SELECT * FROM large_table` without LIMIT → EXPLAIN should detect `ALL` scan type → warn/block depending on setting
- **DDL**: Right-click table → Show DDL → should show `SHOW CREATE TABLE` output
- **Import**: Import a DBeaver `data-sources.json` containing a MySQL connection → should import as `mysql` type with correct host/port
- **Chart**: Open table data → Visualize → select time column with bucket → generated SQL should use `DATE_FORMAT()` not `date_trunc()`
- **MCP**: Use `add_connection` tool with `type: "mysql"` → should create and connect
- **Connection form**: "MySQL / MariaDB" option visible in DB type dropdown with default port 3306

## Checklist
- [x] README updated (tagline, comparison table, safe mode, e2e)
- [x] CHANGELOG updated (Added section)
- [x] CLAUDE.md updated (description, Drivers section, Key Conventions)
- [x] l10n updated (`extension.description`)
- [ ] Wiki: MySQL connection guide needed (not pushed — note only)

---
_Generated by [Claude Code](https://claude.ai/code/session_015RGDJr9CF8pkR6Fk4G7331)_


Siyet (commented):
## Fixes pushed

- `760c451` — `cancelQuery()` now tracks the active connection's `threadId` instead of querying `CONNECTION_ID()` on a fresh pool connection (which always killed its own query, never the target)
- `8168dc9` — Safe mode `scanMatch` for MySQL had no regex capture group, so the warning displayed `Full Table Scan on "undefined"`; fixed with nullish coalescing fallback to `'unknown'`
- `ecc2fda` — Custom time buckets fell through to PostgreSQL `date_bin()` for MySQL; added MySQL-specific branch using `FROM_UNIXTIME(FLOOR(UNIX_TIMESTAMP(col)/N)*N)` with 3 unit tests

## Issues to address

- `src/types/chart.ts:294` — `buildSqliteCustomBucket` hardcodes double-quote column quoting (`"${column}"`). If a future caller passes a MySQL column name it will produce invalid SQL. Consider using the `q()` pattern from `buildAggregationQuery`.
- `src/drivers/mysql.ts:159` — `plan.includes('ALL')` for safe mode full-scan detection is fragile — `ALL` can appear in table/column names or EXPLAIN output values other than the `type` column. A more precise check like matching `" ALL "` (space-delimited) or the column position would reduce false positives.
- `src/commands/queryCommands.ts:153` — MySQL uses `EXPLAIN` but its output format differs from PostgreSQL's text plan. The `plan` string joins `Object.values(row)` with spaces, so extracting the table name from the `table` column position is unreliable. Consider parsing the EXPLAIN result rows directly (the `table` key) instead of regex on the flattened string.

---

PR #120, opened 2026-04-28
feat(mcp): gate add_connection with off/restricted/unrestricted modes

### LINKED ISSUE #119: Standalone MCP: agent can self-issue write-capable connections via add_connection, bypassing all access/write gates
## Context

The standalone MCP server exposes an `add_connection` tool ([`src/mcp-server/index.ts:220-231`](https://github.com/Siyet/viewstor/blob/main/src/mcp-server/index.ts#L220-L231)) that lets an agent create a new database connection and persist it to `~/.viewstor/connections.json`:

```ts
case 'add_connection': {
  const { name, type, host, port, username, password, database, ssl, readonly } = args;
  const id = Date.now().toString(36) + Math.random().toString(36).substring(2, 8);
  const config = { id, name, type, host, port, username, password, database, ssl, readonly };
  await store.add(config);
  return jsonResponse({ id, name, type, host, port, database, message: 'Connection added' });
}
```

`readonly` is taken from the agent-supplied arguments verbatim, and `store.add` writes to the user-scoped config file ([`connectionStore.ts:139-149`](https://github.com/Siyet/viewstor/blob/main/src/mcp-server/connectionStore.ts#L139-L149)).

This is a backdoor around #71 and #75:

1. The user marks all prod connections `agentAccess: 'none'` (per #71). The agent can still `add_connection` a fresh handle to the same host with `readonly: false` and start writing.
2. The user enables `agentWriteApproval: 'always'` (per #75) on every existing connection. The agent's freshly-added connection inherits no profile/approval defaults from the user's folders and falls under whatever the global default ends up being.
3. The agent's connection persists in `~/.viewstor/connections.json` between MCP sessions and shows up in the user's tree on next VS Code reload as if the user added it. Easy to miss in a long connection list.

The same pattern would show up in the in-process surface only if a future tool exposes connection creation; today it is standalone-only, so the fix is contained.

## Threat model

Same-host, same-user agent that has read access to the workspace and the standalone MCP stdio. We are not trying to stop a malicious user — we are stopping an agent from quietly escalating its own privilege when the user thought they had locked things down via #71 / #75.

Concrete sequence the user wants impossible:

1. User configures prod connection `pg-prod` with `readonly: true`, `agentAccess: 'schema-only'`, `agentWriteApproval: 'always'`.
2. Agent reads `getSchema` on `pg-prod`, learns host/port/database.
3. Agent prompts the user (or is prompted) to "set up a working connection for me" with a plausible-looking name `pg-prod-work`, same host, `readonly: false`.
4. From this point on the agent has a fully unrestricted handle to prod.

Step 3 is social-engineering, but the only thing standing between the agent and step 4 is the user reading every `add_connection` call carefully — which is the exact gap the Railway / Cursor incident report flagged.

## Proposed solution

Three options, in increasing strictness. Pick one (or expose as a setting).

### Option A — remove `add_connection` from the standalone MCP

Cleanest. The agent has no business creating new credentials at runtime; the user does that through the VS Code form. The tool was added for convenience but its threat surface outweighs the use case.

Removal is one delete in `setRequestHandler(ListToolsRequestSchema)` plus the case body. Drop the now-unused `add` method on `connectionStore` (or leave it for the in-process side if needed).

### Option B — force-restrict any agent-created connection

If we want to keep the tool, the connection it creates is locked-down by construction:

- `readonly: true` always — the `readonly` argument is rejected (or silently ignored with a warning in the response).
- `scope: 'project'` always — write to `.vscode/viewstor.json` in `process.cwd()`, never to `~/.viewstor/connections.json`. Project-scoped configs strip passwords on save (already true in the VS Code path), so the agent's credential never lands in user globalState.
- `agentAccess: 'schema-only'` and `agentWriteApproval: 'always'` baked in (once #71 / #75 land).
- Connection name prefixed `[agent] ` or similar so the user can spot agent-added entries in the tree.

The agent can still propose a connection for read use (handy for "connect to this docker postgres I just spun up"), but cannot escalate.

### Option C — gate behind a one-time user confirmation

When the agent calls `add_connection`, refuse immediately with a structured error:

```json
{
  "error": "Connection creation requires user confirmation. Open VS Code → Viewstor: New Connection.",
  "kind": "requires_user_action",
  "suggested": { "name": "...", "type": "...", "host": "..." }
}
```

The agent can suggest the connection details, the user creates it manually. No automated path at all. Strictest, but breaks the "agent sets up a sandbox DB for itself" workflow.

### Recommendation

Option B as default + setting `viewstor.standaloneMcp.allowAddConnection` (`'off' | 'restricted' | 'unrestricted'`, default `'restricted'`). `'off'` = Option A behavior, `'unrestricted'` = today's behavior for users who explicitly opt in. Users who want C can use `'off'` and create connections manually.

This composes cleanly with #71 (the new connection inherits the secure-by-default profile from #71's `defaultConnectionProfile`) and #75 (its writes go through the same approval gate even though `readonly: true` already blocks them).

## Out of scope

- Detecting that the agent's new connection points at the *same host* as an existing protected one. Useful, but host-equality on PG is fuzzy (replicas, pooler, alternate ports) and doing it right needs DNS resolution. Follow-up.
- Sandbox / docker auto-detection. We cannot reliably tell prod from local dev from the connection string alone.

## Acceptance criteria

- [ ] `viewstor.standaloneMcp.allowAddConnection` setting (`'off' | 'restricted' | 'unrestricted'`, default `'restricted'`)
- [ ] `'off'` → tool removed from `ListToolsRequestSchema` response and `add_connection` returns `{ error: 'Tool disabled', kind: 'tool_disabled' }`
- [ ] `'restricted'` → connection forced to `readonly: true`, `scope: 'project'`, name prefixed `[agent] `; agent-supplied `readonly: false` ignored with a warning in the response
- [ ] `'unrestricted'` → today's behavior, with a `warnings: [{ kind: 'agent_created_writeable_connection' }]` field in the response
- [ ] Audit log entry (per #75) for every `add_connection` call regardless of mode, including the resolved final config
- [ ] Tree-view indicator for connections created via MCP (a small badge or italic name) so the user can spot agent-added entries
- [ ] Unit tests: each mode × `readonly: true | false | omitted` → expected persisted config
- [ ] Integration test: with `'restricted'`, agent calls `add_connection` then `execute_query` with an UPDATE → returns readonly-blocked error
- [ ] Docs: README + wiki note in the agent-safety section explaining the threat and the three modes

## Related

- #23 — original standalone MCP server (introduced `add_connection`)
- #71 — agent access modes / encrypted credentials (this issue is the bypass route around it)
- #75 — write approval / EXPLAIN safe mode for MCP queries (this issue is the bypass route around it)

Siyet:
## Summary
- Adds `allowAddConnection` setting (`off` / `restricted` / `unrestricted`, default `restricted`) to the standalone MCP server, preventing agents from bypassing per-connection access controls (#71, #75) by self-issuing writeable connections
- `restricted` mode forces `readonly: true`, `scope: 'project'`, name prefixed `[agent] `; `off` hides the tool entirely; `unrestricted` preserves today's behavior with a warning
- Agent-created connections marked `mcpCreated: true` and show `[agent]` badge in the VS Code tree view

Closes #119

## Assumptions
- Setting is read from `~/.viewstor/settings.json` (new file) and env var `VIEWSTOR_ALLOW_ADD_CONNECTION` (takes precedence) since the standalone MCP runs outside VS Code and cannot access `vscode.workspace.getConfiguration()`
- The VS Code `package.json` setting `viewstor.standaloneMcp.allowAddConnection` is registered for discoverability but not wired to the standalone server (the standalone reads its own config)
- Audit log entry (acceptance criteria item) is deferred — #75 has not yet landed the audit infrastructure, so there is no existing log to append to. Once #75 merges, a follow-up can wire `add_connection` events into the same audit path
- Integration test (restricted → execute_query with UPDATE → readonly-blocked) is deferred — it requires a live database and belongs in `test/e2e/`

## Manual test cases
- **Golden path (restricted):** Start standalone MCP (`node dist/mcp-server.js`), call `add_connection` with `readonly: false` → verify response shows `readonly: true`, `scope: 'project'`, name starts with `[agent] `, `warnings` includes `readonly_forced`; verify `.vscode/viewstor.json` was created with password stripped
- **Off mode:** Set `VIEWSTOR_ALLOW_ADD_CONNECTION=off`, restart MCP → call `list_tools` → verify `add_connection` is absent; call `add_connection` directly → verify error with `kind: 'tool_disabled'`
- **Unrestricted mode:** Set `VIEWSTOR_ALLOW_ADD_CONNECTION=unrestricted` → call `add_connection` with `readonly: false` → verify response preserves `readonly: false`, saves to `~/.viewstor/connections.json`, includes `agent_created_writeable_connection` warning
- **Tree badge:** Open VS Code after a restricted `add_connection` → verify the connection shows `[agent] host:port` in the tree description

## Checklist
- [x] README — added `add_connection` security note in standalone MCP section
- [x] CHANGELOG — added entry under `[Unreleased] > Added`
- [x] CLAUDE.md — documented `settings.ts`, `addProjectScoped()`, and `add_connection` gating
- [x] l10n — added 4 strings for the new VS Code setting
- [ ] Wiki — `add_connection` security section (note: wiki edits cannot be pushed via PR)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FKjxFZLo5fcLXv9pfwem4S)_


Siyet (commented):
## Fixes pushed
- `2cfb026` — Include `kind: 'tool_disabled'` in off-mode error response. Acceptance criteria #2 requires `{ error: '...', kind: 'tool_disabled' }` but the handler was using the generic `errorResponse()` helper which only emits `{ error: message }`. The test's mock function returns `kind` correctly, so the mock and the real handler now match.

## Issues to address
- `src/mcp-server/index.ts:230-233` — `host` and `port` are typed as required (`string`/`number`) in the destructuring cast, but the input schema only requires `['name', 'type']`. For SQLite connections where host/port are irrelevant, the agent may omit them, leaving `undefined` at runtime despite the type assertion. Consider making them optional in the cast or adding defaults. (Pre-existing pattern, not a regression from this PR, but worth fixing since the PR already touches this code path.)


---

PR #123, opened 2026-04-29
fix(mcp): gate add_connection behind mode setting (#119)

### LINKED ISSUE #119: Standalone MCP: agent can self-issue write-capable connections via add_connection, bypassing all access/write gates

Siyet:
## Summary
- Add `viewstor.standaloneMcp.allowAddConnection` setting (`off` / `restricted` / `unrestricted`, default `restricted`) to prevent agents from self-issuing writeable connections that bypass `readonly`, `agentAccess`, and `agentWriteApproval` controls
- In restricted mode: force `readonly: true`, `scope: 'project'`, name prefixed `[agent] `, passwords stripped from persisted project config
- Audit log entry written to `~/.viewstor/audit.log` on every `add_connection` call; tree view shows `(agent)` badge on agent-created connections

Closes #119

## Assumptions
- Setting is read from env var `VIEWSTOR_ALLOW_ADD_CONNECTION` or `~/.viewstor/config.json` (under `standaloneMcp.allowAddConnection`) since the standalone MCP server runs outside VS Code and cannot access VS Code settings directly. The VS Code setting is registered in `package.json` for documentation/discoverability.
- Integration test for restricted-mode execute_query blocking UPDATE is deferred — the standalone MCP server requires a live stdio transport and database connection. The unit tests cover all mode × readonly combinations and the project-scoped save behavior.
- Tree-view indicator uses `(agent)` text in the description rather than a custom icon, matching the existing description pattern for host:port.

## Manual test cases
- **Golden path (restricted mode):** Start standalone MCP server with no env override → call `list_tools` → `add_connection` is listed → call `add_connection` with `name: "Prod PG"`, `type: "postgresql"`, `host: "db.prod"`, `port: 5432`, `readonly: false` → response contains `name: "[agent] Prod PG"`, `readonly: true`, warning `readonly_forced`, message mentions "restricted mode" → check `.vscode/viewstor.json` contains the connection without password → check `~/.viewstor/audit.log` has an entry
- **Off mode:** Set `VIEWSTOR_ALLOW_ADD_CONNECTION=off` → call `list_tools` → `add_connection` not in tool list → call `add_connection` → returns error "Tool disabled"
- **Unrestricted mode:** Set `VIEWSTOR_ALLOW_ADD_CONNECTION=unrestricted` → call `add_connection` with `readonly: false` → response contains `warnings: [{ kind: 'agent_created_writeable_connection' }]`, name is NOT prefixed
- **Tree badge:** Open VS Code with a project-scoped connection that has `agentCreated: true` → tree shows `(agent)` next to the host:port description

## Checklist
- [x] README — added agent connection safety paragraph in MCP section
- [x] CHANGELOG — added entry under [Unreleased]
- [x] CLAUDE.md — updated Standalone MCP Server section with mode documentation
- [x] l10n — added setting description and enum descriptions
- [ ] Wiki — note in PR body: update MCP Server wiki page with the three modes and threat model

---
_Generated by [Claude Code](https://claude.ai/code/session_01HVssLcY4ZNSARHgwrLMHoR)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to Viewstor are documented here. Format based on [Keep a Cha
 
 ## [Unreleased]
 
+### Fixed
+- **Standalone MCP `add_connection` bypasses access/write gates** — agents could self-issue writeable connections to the same host, circumventing `readonly`, `agentAccess`, and `agentWriteApproval` settings. New `viewstor.standaloneMcp.allowAddConnection` setting (`off` / `restricted` / `unrestricted`, default `restricted`): in restricted mode agent-created connections are forced read-only, project-scoped (`.vscode/viewstor.json`, passwords stripped), and name-prefixed `[agent] `; in off mode the tool is removed from the tool list entirely. Audit log entry written to `~/.viewstor/audit.log` on every call. Tree view shows `(agent)` badge on agent-created connections. Configurable via `~/.viewstor/config.json` or `VIEWSTOR_ALLOW_ADD_CONNECTION` env var ([#119](https://github.com/Siyet/viewstor/issues/119))
+
 ## [0.4.0] — 2026-04-29
 
 ### Fixed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -175,9 +175,16 @@ All auto-connect. Returns structured JSON or `{ error }`.
 ### Standalone MCP Server
 `src/mcp-server/index.ts` — stdio-based MCP server for CLI agents (Claude Code, etc.). Built as separate webpack entry → `dist/mcp-server.js`. Uses `@modelcontextprotocol/sdk`. Does NOT import `vscode`.
 
-`src/mcp-server/connectionStore.ts` — reads connections from `~/.viewstor/connections.json` (user) and `.vscode/viewstor.json` (project). Manages driver lifecycle.
+`src/mcp-server/connectionStore.ts` — reads connections from `~/.viewstor/connections.json` (user) and `.vscode/viewstor.json` (project). Manages driver lifecycle. `getAddConnectionMode()` reads `viewstor.standaloneMcp.allowAddConnection` from env var `VIEWSTOR_ALLOW_ADD_CONNECTION` or `~/.viewstor/config.json` (`standaloneMcp.allowAddConnection`), defaulting to `'restricted'`.
 
-9 tools: `list_connections`, `get_schema`, `execute_query`, `get_table_data`, `get_table_info`, `add_connection`, `reload_connections`, `build_chart`, `export_grafana_dashboard`.
+9 tools: `list_connections`, `get_schema`, `execute_query`, `get_table_data`, `get_table_info`, `add_connection` (gated by mode), `reload_connections`, `build_chart`, `export_grafana_dashboard`.
+
+`add_connection` mode (`'off' | 'restricted' | 'unrestricted'`, default `'restricted'`):
+- `off` → tool removed from `ListToolsRequestSchema`; calling returns `{ error: 'Tool disabled' }`.
+- `restricted` → connection forced to `readonly: true`, `scope: 'project'`, name prefixed `[agent] `, `agentCreated: true`. Agent-supplied `readonly: false` is silently overridden with a warning in the response.
+- `unrestricted` → today's behavior + `warnings: [{ kind: 'agent_created_writeable_connection' }]` when the connection is writeable.
+
+Every `add_connection` call writes an audit log entry to `~/.viewstor/audit.log` with timestamp, mode, and the resolved config (password stripped).
 
 Data-oriented tools (`execute_query`, `get_schema`, `get_table_data`, `get_table_info`, `build_chart`) accept an optional `database` parameter. `ConnectionStore.ensureDriverForDatabase()` mirrors `ConnectionManager.getDriverForDatabase()` — caches per `connectionId:database`, reuses host/user/password/ssl. VS Code MCP commands accept `database` as a trailing optional arg.
 
--- a/README.md
+++ b/README.md
@@ -226,6 +226,8 @@ Or add manually:
 
 All MCP interfaces auto-connect and respect read-only mode.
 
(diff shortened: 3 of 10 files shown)


Siyet (commented):
## Fixes pushed

- `af333a0` — Add `kind: 'tool_disabled'` to off-mode error response per issue #119 acceptance criteria; fix `saveProjectConfig` dedup filter; update test.

## Issues to address

- `src/test/addConnectionMode.test.ts:67-103` — The `add_connection mode behavior` tests use a local `simulateAddConnection` function that re-implements the handler logic rather than testing the actual `index.ts` handler. If the handler diverges from the simulation, the tests won't catch it. Consider extracting the mode-handling logic from the handler into a testable pure function in `connectionStore.ts`.

---

